PR #8, opened 2020-04-10
Use gulrak instead of filesystem, lower req to C++ 11

mariano54:
I have not confirmed that it works with Mojave, but this is what the compile options were like before filesystem was added.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,6 @@
 name: Build
 
+#on: [push, pull_request]
 on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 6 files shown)


mariano54:
Do you know why it fails on windows @wjblanke ?

hoffmang9:
image

wjblanke:
we aren't using mingw anymore. you could try testing with visual studio and awjb branch

hoffmang9:
But it did kill the mingw64 build for some reason.

wjblanke:
Yup

PR #39, opened 2020-06-09
Add target_link_libraries configuration for OpenBSD

n1000:
OpenBSD uses clang / clang++ by default and does not need the stdc++fs dependency.


hoffmang9:
Thanks for the PR! I will go ahead and merge but we just had a full release earlier today so it will be a week or three before this makes it into the next release.

PR #40, opened 2020-06-12
Add target_link_libraries configuration for FreeBSD

n1000:
Tested on FreeBSD amd64 11.3 and 12.1.


hoffmang9:
Note that I just made some changes which you may want to rebase against so that ci will work on your pull requests.

PR #50, opened 2020-07-31
Optimize rmap vector in FindMatches

rostislav:
Replace inner vectors with a small struct.


rostislav:
Replaced with a new PR

PR #56, opened 2020-08-14
Cpu opt

hoffmang9:


hoffmang9:
I'm running a test k30 on EC2 memory plotting. If that matches hashes from a 2nd test k30 plot of same index - any reason not to bring this on in to chia-blockchain for release @rostislav ?

rostislav:
Sure, you can include these changes into the release, I consider them low-risk

hoffmang9:
Also thanks for knocking down those LGTM issues!

PR #58, opened 2020-08-31
Ms.metadata

mariano54:
Implements rewriting of left tables during forward prop, to save working space. Also drops unused entries in forward prop. 
Overall seems to save 40-55%  in working space. Reduces total writes to ssd by up to 20%.

* Tmp files now separated into one table per file
* plot pointers removed, replaced with a vector of FileDisks
* Files truncated when stuff is not needed
* Fixed an edge case with reading and writing zeroes to the end of the tables
* Backprop slightly sped up by dropping entries, and sorting on more uniform data
* Positions stored in k bits instead of k+1
* Fix valgrind memory leaks/errors by Freeing ANS resources, and initializing memory

Slightly slows down overall plotting by around 3% due to forward prop CPU usage. Some optimizations still left to do here.

### DIFF
--- a/src/plotter_disk.hpp
+++ b/src/plotter_disk.hpp
@@ -267,7 +267,11 @@ class DiskPlotter {
         switch (table_index) {
             case 1:
                // Represents f1, x
-               return Util::ByteAlign(k + kExtraBits + k) / 8;
+               if (phase_1_size) {
+                    return Util::ByteAlign(k + kExtraBits + k) / 8;
+               } else {
+                    return Util::ByteAlign(k) / 8;
+               }
             case 2:
             case 3:
             case 4:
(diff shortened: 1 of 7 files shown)


hoffmang9:
`./ProofOfSpace -k 30 -t /mnt/plot1/ -2 /mnt/plots/ -d /mnt/plots generate > k30-ms.log &`


```
operation: generate
Generating plot for k=30 filename=plot.dat id=022fb42c08c12de3a6af053880199806532e79515f94e83461612101f9412f9e


Starting plotting progress into temporary dirs: /mnt/plot1/ and /mnt/plots/
Memo: 0102030405
ID: 022fb42c08c12de3a6af053880199806532e79515f94e83461612101f9412f9e
Plot size is: 30
Buffer size is: 2147483648

Starting phase 1/4: Forward Propagation into table tmp files... Mon Aug 31 19:23:05 2020
Computing table 1
F1 complete, Time =  99.719 seconds. CPU (99.99%) Mon Aug 31 19:24:45 2020
Computing table 2
        Sorting table 1
Only read 2 of 9 bytes at offset 2066640894 from "/mnt/plot1/plot.dat.sort.tmp"with length 9663676416. Error 0. Retrying in five minutes.
```

/mnt/plot1 and /mnt/plots are NVMe on ec2.

mariano54:
Thanks for the report, I'll see what's going on. 
@wjblanke can you look through the code changes to see if anything looks off to you in terms of reading/writing to disk?

hoffmang9:
Proof verification suceeded. k = 32
Total success: 979/1000, 97.9%.

k=32 used a maximum of 317 GiB

k=30 testing showed mild speed increases.

hoffmang9 (approved):
after that final hairy windows bug, looks good to me.

src/plotter_disk.hpp
```diff
@@ -146,35 +166,41 @@ class DiskPlotter {
             assert(id_len == kIdLen);
 
             // Memory to be used for sorting and buffers
-            uint8_t* memory = new uint8_t[memorySize];
```
arvidn, line 149: @mariano54 is there a good reason to clear the memory here? this adds a noticeable delay on startup

arvidn: is the idea to pre-fault the memory into physical RAM and the TLB? I think that would only help with latency later, not necessarily increase throughput

PR #59, opened 2020-09-04
CPU optimizations and minor fixes to metadata optimizations

rostislav:
This includes all my earlier CPU optimizations. The last two commits are minor fixes to Mariano's metadata optimizations.


mariano54:
Has the code to add FileDisk s in a loop been tested? when I tried this, it was calling the destructor and thus closing the file. So please ensure this is not happening.

hoffmang9:
I'm speed testing it under Ubuntu 20.04 and will install it and test plot on Windows shortly.

PR #63, opened 2020-09-24
Stripewjb

wjblanke:
This implements the striping which is the most complicated part of ramwjb2. It should work for spinny disks too. I wanted to let Mariano take a look at it, especially as I hope to merge it before his PR. It is set to 1 CPU in the code. Change NUMTHREADS to what u please.


mariano54:
Let's add a command line argument for number of threads

src/plotter_disk.hpp
```diff
@@ -73,8 +76,639 @@ struct Phase3Results {
...
+{
+    size_t i = 0;
+    for (i = 0; i < buf_len; ++i)
+        fprintf(stdout, "%02X%s", buf[i], (i + 1) % 16 == 0 ? "\r\n" : " ");
+}
+
 const Bits empty_bits;
 
+#define STRIPESIZE 8192
+#define NUMTHREADS 2
```
arvidn, line 89: these days it's preferred to use `const`:
```
int const stripesize = 8192;
int const numthreads = 1;
```

src/plotter_disk.hpp
```diff
@@ -73,8 +76,639 @@ struct Phase3Results {
...
+
+typedef struct {
```
arvidn, line 91: in C++ you don't need to create typedefs for classes, their names are proper types. You can just do:

```
struct THREADDATA
{
   // ...
};
```

src/plotter_disk.hpp
```diff
@@ -16,8 +16,11 @@
 #define SRC_CPP_PLOTTER_DISK_HPP_
 
 #ifndef _WIN32
+#include <pthread.h>
```
arvidn, line 19: thread support is standarized. you can include `<thread>` and have a portable `std::thread` class

arvidn:
this branch fails one of the tests for me, when I run it locally on Mac:

```
Starting phase 1/4: Forward Propagation into tmp files... Tue Oct  6 12:14:50 2020
Computing table 1
Assertion failed: (size == 64 || value == (value & ((1ULL << size) - 1))), function InitBitsGeneric, file src/bits.hpp, line 138.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
RunTests is a Catch v2.5.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
Plotting
  Disk plot 1
-------------------------------------------------------------------------------
tests/test.cpp:542
...............................................................................

tests/test.cpp:542: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```

src/plotter_disk.hpp
```diff
@@ -73,8 +76,639 @@ struct Phase3Results {
...
+        if (pos < prevtableentries + 1) {
+            uint64_t readamt = std::min(
+                ((uint64_t)(STRIPESIZE) + 2500) * entry_size_bytes,
+                ((prevtableentries + 1) * entry_size_bytes) - left_reader);
+
+            (*ptmp_1_disks)[table_index].Read(left_reader, left_reader_buf, readamt);
+            left_reader_count = 0;
+        }
+#ifdef _WIN32
+        ReleaseSemaphore(ptd->mine, 1, NULL);
```
arvidn, line 378: it's not clear to me why the threads wait for each other (in order), and then perform some serial work, and then kick the next thread to do some serial work. You have this cascade of semaphores to implement this.

Wouldn't it be simpler if all the parallel work happens first, then all threads join and then all the serial work is performed? Is the serial work so expensive there's a material gain from allowing threads ahead of it to keep working on the parallel work while one thread performs the serial work?

wjblanke: Hi arvid

Yes we are cascading semaphores here to order things. The workload is divided into small stripes as we don't have enough memory to do the whole thing. The stripesize is configurable though, so if you had enough memory you could divide it evenly as you have said.  We have to serialize things because the stripe can only calculate the relative position within the stripe but the table demands the absolute position. So we have to correct things serially as the table is written out. There are other reasons for serializing things as well (limiting random disk seeks eg). Currently the code does not scale up very well past a few CPUs though so certainly lots of room for improvement here. The entry dropping code demands a lot of synchronization but it is worth it for the typical user. ramwjb runs fastest (48 cpu load) but all ram is pretty specialized. Let me know if you have any ideas!!!!

wjblanke: BTW there is a video in keybase about the entry dropping and striping: (striping-bill)

arvidn: I see, what I had missed was that the thread continues the "parallel work" again. I first had read it as it was doing the serial work at the end, right before the thread would finish, and return from the thread function.

PR #64, opened 2020-09-24
Ms.fast sort++

mariano54:
>70% decreased writes to disk
>25-50% increased speed 

* Adds a new sort on disk algorithm: fast_sort_on_disk.hpp. Before, we were writing out all entries naively, and then sorting. Now, we write out entries into 16 different files, depending on the bucket, or first 4 bits of the entry. Then we sort each one, as we read them in the next table. This saves total two write passes and two read passes over each table.
* Recursive sort has been removed
* Dynamically sets the number of buckets so that an entire bucket fits in memory. Num buckets can vary from 16 to 128. The small one is preferred, since it reduces seeks on disk. For HDDs, more memory is important, so that we can keep the bucket size low.
* Due to these limitations, there is a now a minimum buffer size.
* Fixes the memory spike issue
* Fixes a bug in compression which dropped some entries, this changes the final file slightly (everything is still compatible)
* Use quicksort if we don't have enough memory for SortInMemory (which now is called UniformSort).
* Divided all the sort and disk objects into separate files (disk.hpp, uniformsort.hpp, quicksort.hpp.hpp, fast_sort_on_disk.hpp).


mariano54:
Close in favor of https://github.com/Chia-Network/chiapos/pull/65

PR #65, opened 2020-09-28
Phase 1 multithreading and fast sort

mariano54:
Combination of stripeswjb and ms.fastSort++.

>70% decreased writes to disk
>25-50% increased speed 

* Enables multithreading for phase 1
* Adds new arguments: u: number of buckets, h: number of threads, s: stripe size for forward prop
* Adds a new sort on disk algorithm: sort_manager.hpp. Before, we were writing out all entries naively, and then sorting. Now, we write out entries into 16 different files, depending on the bucket, or first 4 bits of the entry. Then we sort each one, as we read them in the next table. This saves total two write passes and two read passes over each table.
* Recursive sort has been removed
* Dynamically sets the number of buckets so that an entire bucket fits in memory. Num buckets can vary from 16 to 128. The small one is preferred, since it reduces seeks on disk. For HDDs, more memory is important, so that we can keep the bucket size low.
* Due to these limitations, there is a now a minimum buffer size.
* Fixes the memory spike issue
* Fixes a bug in compression which dropped some entries, this changes the final file slightly (everything is still compatible)
* Use quicksort if we don't have enough memory for SortInMemory (which now is called UniformSort).
* Divided all the sort and disk objects into separate files (disk.hpp, uniformsort.hpp, quicksort.hpp.hpp, fast_sort_on_disk.hpp).
* Splits phases into their own files
* reduce seeks in WriteParkToFile (credit to xorinox for finding the issue)
* remove logging of memo which contains plot private key
* Does not use more memory than what is passed in with -b


src/cli.cpp
```diff
@@ -68,18 +68,24 @@ int main(int argc, char *argv[])
...
         string memo = "0102030405";
         string id = "022fb42c08c12de3a6af053880199806532e79515f94e83461612101f9412f9e";
-        uint32_t buffmegabytes = 2 * 1024;  // 2 gigabytes
+        uint32_t buffmegabytes = 0;
 
         options.allow_unrecognised_options().add_options()(
-            "k, size", "Plot size", cxxopts::value<uint8_t>(k))(
-            "t, tempdir", "Temporary directory", cxxopts::value<string>(tempdir))(
+                "k, size", "Plot size", cxxopts::value<uint8_t>(k))(
+                "h, threads", "Number of stripes", cxxopts::value<uint8_t>(num_threads))(
```
hoffmang9, line 85: Typo - should be Number of threads

src/cli.cpp
```diff
@@ -68,18 +68,24 @@ int main(int argc, char *argv[])
...
+                    "u, buckets", "Number of buckets", cxxopts::value<uint32_t>(num_buckets))(
+                "s, stripes", "Number of stripes", cxxopts::value<uint32_t>(num_stripes))(
```
hoffmang9, line 87: Is this number of or size of?

hoffmang9 (changes requested):
With those two typo's to fix LGTM

hoffmang9 (approved):
Now LGTM fully

PR #66, opened 2020-09-29
Increase default stripe size

mariano54:


hoffmang9 (approved):
LGTM and RunTests seems happier

PR #68, opened 2020-09-29
Ms.race

mariano54:


src/phase1.hpp
```diff
@@ -224,15 +226,18 @@ void* phase1_thread(void* arg)
             stripe_start_correction = 0;
         }
 
-        bool need_new_bucket = globals.L_sort_manager->CloseToNewBucket(left_reader);
-
+        SemaphoreUtils::Wait(ptd->theirs);
```
wjblanke, line 229: Why are we calling wait twice on the same semaphore?

hoffmang9 (approved):
LGTM

PR #70, opened 2020-10-02
Simplify and improve performance of F1 calculation

hoffmang9:
Do not use Bits, provide F1 values in a simple array instead of a vector
of Bits. This increases performance of F1 calculation by several times.

### DIFF
--- a/src/calculate_bucket.hpp
+++ b/src/calculate_bucket.hpp
@@ -77,7 +77,9 @@ public:
     inline F1Calculator(uint8_t k, const uint8_t* orig_key)
     {
         uint8_t enc_key[32];
+        size_t buf_blocks = CDIV((k + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2;
         this->k_ = k;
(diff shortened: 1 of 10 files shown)


mariano54 (approved):

src/calculate_bucket.hpp
```diff
@@ -77,7 +77,9 @@ class F1Calculator {
     inline F1Calculator(uint8_t k, const uint8_t* orig_key)
     {
         uint8_t enc_key[32];
+        size_t buf_blocks = CDIV((k + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2;
```
arvidn, line 80: note that my patch of making this a function (and spell lower case `cdiv()`) has landed in `master` already. So this would need to change before landing.

You may want to rebase the patch on top of master to catch any other potential issues like this

hoffmang9:
Closing in favor of https://github.com/Chia-Network/chiapos/pull/83

PR #75, opened 2020-10-05
remove unused stack_allocator

arvidn:
Was this added with the intention of being used in the future? a cursory look through the history didn't reveal any obvious use of it ever


hoffmang9:
Hellman attacks were there for the origin of this repo which was the proof of space competition. We do want the tools around so that people can attempt them but there is an amorphous direction to take the Hellman attacks into a new repository

hoffmang9 (approved):
LGTM

arvidn:
as far as I can tell, `stack_allocator` wasn't even used in `hellman_example` code.

PR #76, opened 2020-10-05
pull in uint128_t as a submodule

arvidn:
I noticed there weren't any unit tests for the uint128_t library. but when I went looking, it turned out the source repository does have tests. This is mostly just a "janitorial" suggestion of pulling in this library as a submodule instead if copy-pasting it into the repo. This way we get the tests (although they aren't hooked up to CI), and make it easier to update to newer versions.

The patch removes the uint128 subdirectory from root and adds uint128_t as a submodule to get the latest version as well as unit tests.

It also fix one small build issue when using the uint128_t library, instead of the built-in type.


arvidn:
I see

hoffmang9:
Just capturing the conversation around this one. uint128 is only used by Windows so cmake fetching it is not an appealing option. However @justinengland will need to figure out a better way to pull this in for the Windows python wheel build.

arvidn:
would it be unreasonable to build the windows python module with cmake as well? (I don't have great experience with setup tools and would generally think it would be desirable to move off of it)

hoffmang9:
I'm indifferent because both paths (cmake or setuptools) are almost equally painful - but cmake wins that nasty race to the bottom... The flip side is that cmake everywhere may be easier to maintain. We will want to roll whichever plan through all the sub repos though.

PR #78, opened 2020-10-06
remove using namespace std; from header file

arvidn:
pulling in all of `std` in a header causes name clashes and generally makes code more brittle. This was causing name clashes in a different patch I'm working on.

### DIFF
--- a/src/bits.hpp
+++ b/src/bits.hpp
@@ -24,8 +24,6 @@
 #include "./util.hpp"
 #include "exceptions.hpp"
 
-using namespace std;
-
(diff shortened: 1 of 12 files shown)


src/bits.hpp
```diff
@@ -24,8 +24,6 @@
 #include "./util.hpp"
 #include "exceptions.hpp"
 
-using namespace std;
```
arvidn, line 27: this is the material change. All other changes fall out from this

hoffmang9 (approved):
lgtm

PR #81, opened 2020-10-06
use signed integer as loop variable for bucket_index

arvidn:
we don't need modulo-2 arithmetic on them.


hoffmang9 (approved):
Looks good

PR #82, opened 2020-10-06
add additional unit tests for the Bits class with a few minor fixes

arvidn:


src/bits.hpp
```diff
@@ -261,12 +261,17 @@ class BitsGeneric {
 
     BitsGeneric<T> Rotl(uint32_t shift) const
```
arvidn, line 262: it turns out this function doesn't seem to be used anywhere, so I imagine it would be better to remove it than to add unit tests to it

hoffmang9: I agree with removing it.

hoffmang9 (approved):
looks good

PR #83, opened 2020-10-06
Simplify and improve performance of F1 calculation

hoffmang9:
This is a cherry-pick rebase of https://github.com/Chia-Network/chiapos/pull/70
It also changes ProofOfSpace.exe to use create instead of generate to match python

### DIFF
--- a/src/calculate_bucket.hpp
+++ b/src/calculate_bucket.hpp
@@ -77,7 +77,7 @@ public:
     inline F1Calculator(uint8_t k, const uint8_t* orig_key)
     {
         uint8_t enc_key[32];
-        size_t buf_blocks = CDIV((k + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2;
+        size_t buf_blocks = cdiv((k + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2;
         this->k_ = k;
         this->buf_ = (uint8_t *)malloc(buf_blocks * kF1BlockSizeBits / 8);
 
--- a/src/calculate_bucket.hpp
+++ b/src/calculate_bucket.hpp
@@ -77,7 +77,9 @@ public:
     inline F1Calculator(uint8_t k, const uint8_t* orig_key)
     {
         uint8_t enc_key[32];
+        size_t buf_blocks = CDIV((k + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2;
         this->k_ = k;
+        this->buf_ = (uint8_t *)malloc(buf_blocks * kF1BlockSizeBits / 8);
 
         // First byte is 1, the index of this table
         enc_key[0] = 1;
@@ -87,7 +89,10 @@ public:
         chacha8_keysetup(&this->enc_ctx_, enc_key, 256, NULL);
     }
 
-    inline ~F1Calculator() {}
+    inline ~F1Calculator()
+    {
+        free(this->buf_);
+    }
 
     // Disable copying
     F1Calculator(const F1Calculator&) = delete;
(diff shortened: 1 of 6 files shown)


src/phase1.hpp
```diff
@@ -601,31 +602,27 @@ std::vector<uint64_t> RunPhase1(
...
 
     // Instead of computing f1(1), f1(2), etc, for each x, we compute them in batches
     // to increase CPU efficency.
-    for (uint64_t lp = 0; lp <= (((uint64_t)1) << (k - kBatchSizes)); lp++) {
-        // For each pair x, y in the batch
-        for (auto kv : f1.CalculateBuckets(Bits(x, k), 2 << (kBatchSizes - 1))) {
-            Bits to_write = std::get<0>(kv) + std::get<1>(kv);
+    for (uint64_t lp = 0; lp < (uint64_t)1 << (k - kBatchSizes); lp++) {
+        f1.CalculateBuckets(x, 1 << kBatchSizes, f1_entries);
+        for (int i = 0; i < 1 << kBatchSizes; i++) {
```
wjblanke, line 613: should use uint_t type

src/phase1.hpp
```diff
@@ -588,6 +588,7 @@ std::vector<uint64_t> RunPhase1(
     F1Calculator f1(k, id);
     uint64_t x = 0;
     uint64_t prevtableentries = 0;
+    uint64_t *f1_entries = (uint64_t *)malloc((1 << kBatchSizes) * sizeof(*f1_entries));
```
wjblanke, line 591: Should check for null

wjblanke (approved):
Looks fine, I added some picky things but could be merged as is.

src/phase1.hpp
arvidn: I think `int` is reasonable. We don't want (or need) modulo-2 arithmetic on this variable

src/calculate_bucket.hpp
```diff
@@ -77,7 +77,9 @@ class F1Calculator {
     inline F1Calculator(uint8_t k, const uint8_t* orig_key)
     {
         uint8_t enc_key[32];
+        size_t buf_blocks = cdiv((k + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2;
         this->k_ = k;
+        this->buf_ = (uint8_t *)malloc(buf_blocks * kF1BlockSizeBits / 8);
```
arvidn, line 82: if you use `new uint8_t[buf_blocks * kF1BlockSizeBits / 8]` instead, you don't need to check for null. But then you'd need to replace `free(this->buf_);` with `delete[] this->buf_;`

Although, ideally, `buf_` would be of type `std::unique_ptr<uint8_t[]>` and you wouldn't call `free()` or `delete[]` at all. I intend to propose a change like that in other places too.

src/calculate_bucket.hpp
```diff
@@ -154,69 +159,25 @@ class F1Calculator {
...
-        const Bits& start_L,
-        uint64_t number_of_evaluations)
+    // F1(x) values for x in range [first_x, first_x + n) are placed in res[].
+    // n must not be more than 1 << kBatchSizes.
+    void CalculateBuckets(uint64_t first_x, uint64_t n, uint64_t *res)
     {
-        uint16_t num_output_bits = k_;
-        uint16_t block_size_bits = kF1BlockSizeBits;
+        uint64_t start = first_x * k_ / kF1BlockSizeBits;
+        uint64_t end = ((first_x + n + 1) * k_ - 1) / kF1BlockSizeBits;
```
arvidn, line 167: there's a pretty well established convention (at least in C++) that "end" refers to the one-past-end position. So to calculate the *number* of entries, you can just do `end - begin`. As far as I can tell, this is the *last* position.

As someone new to the conventions in this code base, it's not all that clear to me what this calculation is doing. The comment above states that `n` is indeed the number of values to be computed, and written to `res`. `end` is only used to compute `n_blks`, which is the number of chacha8 keys to be written to `buf_`, each 64 bytes (`kF1BlockSizeBits` bits).

The intuitive way of computing `end` (I think) would be: `(first_x + n) * k_ / kF1BlockSizeBits`. It looks like what you're doing is a kind of rounding-up, to `k_`, not to `kF1BlockSizeBits`.

in short, I think this computation could use a comment, and either make `end` be one-past-end, or renamed `last`

src/calculate_bucket.hpp
```diff
@@ -154,69 +159,25 @@ class F1Calculator {
...
-        // Counter for the last input
-        uint64_t counter_end =
-            ((start_L.GetValue() + (uint128_t)number_of_evaluations + 1) * num_output_bits) /
-            block_size_bits;
-
-        std::vector<Bits> blocks;
-        uint64_t L = (counter * block_size_bits) / num_output_bits;
-        uint8_t ciphertext_bytes[kF1BlockSizeBits / 8];
+        chacha8_get_keystream(&this->enc_ctx_, start, n_blks, buf_);
+        for (uint64_t i = 0; i < n; i++) {
```
arvidn, line 173: it may be a further simplification to use `x` as the loop variable directly:

```
for (uint64_t x = first_x; x < first_x + n; ++x) {
```

since both `first_x` and `n` are constants the loop condition is cheap. (I think you should mark all the constants as `const` too though, to make it more obvious to readers)

src/calculate_bucket.hpp
```diff
@@ -154,69 +159,25 @@ class F1Calculator {
...
-            chacha8_get_keystream(&this->enc_ctx_, counter, 1, ciphertext_bytes);
-            Bits ciphertext(ciphertext_bytes, block_size_bits / 8, block_size_bits);
-            blocks.push_back(std::move(ciphertext));
-            ++counter;
-        }
-
-        std::vector<std::pair<Bits, Bits>> results;
-        uint64_t block_number = 0;
-        uint16_t start_bit = (start_L.GetValue() * (uint128_t)num_output_bits) % block_size_bits;
+            res[i] = (y << kExtraBits) | (x >> x_shift);
```
arvidn, line 177: there is an assumption that `_k + kExtraBits <= 64`, right? Otherwise bits from `y` would be lost.

rostislav: Yes, that's correct. Minimum and maximum `k_` values are defined by `kMinPlotSize` and `kMaxPlotSize`, which are 18 and 50. `kExtraBits` is 6.

hoffmang9:
My k=32 test plot completed successfully.

F1 complete, time: 233.849 seconds. CPU (76.54%)
Approximate working space used (without final file): 286.580 GiB
Final File size: 101.326 GiB
Total time = 82625.894 seconds. CPU (50.220%)

Older version for comparison:
F1 complete, time: 469.209 seconds. CPU (99.99%)

src/phase1.hpp
wjblanke: oops sorry i meant we should always specify the width, ala uint64_t, instead of using generic int which is platform specific

arvidn: makes sense. my concern, generally, with unsigned types is that all arithmetic is effectively two operations. the arithmetic operation as well as an `& 0xffffffff` (or whatever width the variable has). As long as your hardware has arithmetic operators with a built-in width (say, ways of referring to smaller parts of registers as x86 does) that's fine. But if it doesn't, if all registers are 64 bits wide, then the compiler will need to lower it to two instructions, which is wasteful where we don't rely on modulo-2 arithmetics

rostislav:
@wjblanke I think my optimization should be in a separate pull request from your implementation, which adds multithreading, since those are distinct features and they can work independently. If it's okay with you, I'll create a separate PR with only my optimization on top of master, with fixes based on Arvid's feedback.

hoffmang9:
Closing in favor of already merged https://github.com/Chia-Network/chiapos/pull/96

PR #84, opened 2020-10-06
use std::thread to simplify code

arvidn:
a simplification to unify posix and windows code by using `std::thread`


hoffmang9 (approved):
With fingers crossed about Windows - looks good.

PR #87, opened 2020-10-06
remove unused functions on the Bits class

arvidn:
that are not used at all or only used by the unit test


tests/test.cpp
```diff
@@ -630,24 +427,24 @@ TEST_CASE("F functions")
...
                 uint64_t bucket = std::get<0>(pair).GetValue() / kBC;
                 if (buckets.find(bucket) == buckets.end()) {
                     buckets[bucket] = vector<std::pair<Bits, Bits>>();
                 }
                 buckets[bucket].push_back(pair);
-                if (x.GetValue() + 1 > (1ULL << k) - 1) {
+                if (x + 1 > (1ULL << k) - 1) {
                     break;
                 }
                 ++x;
```
arvidn, line 445: this was the only place where the increment operator was actually used (and not just tested). However, this code just uses `Bits x` as a `uint64_t` already anyway, so I think the new version is still simpler

hoffmang9 (approved):
lgtm

PR #91, opened 2020-10-07
CLI ergonomics

arvidn:
There's a large indentation change, so it's best reviewed ignoring whitespace.

### DIFF
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -63,212 +63,207 @@ void HelpAndQuit(cxxopts::Options options)
     exit(0);
 }
 
-int main(int argc, char *argv[])
-{
-    try {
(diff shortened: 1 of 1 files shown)


src/cli.cpp
```diff
@@ -114,17 +114,21 @@ int main(int argc, char *argv[])
             cout << "Generating plot for k=" << static_cast<int>(k) << " filename=" << filename
                  << " id=" << id << endl
                  << endl;
+            id = Strip0x(id);
```
arvidn: if there is a `0x` prefix, it should be stripped *before* checking that the length is 64 characters

hoffmang9 (approved):
Looks good

PR #94, opened 2020-10-09
Cli create, deprecate pep517.build

hoffmang9:


src/cli.cpp
```diff
@@ -56,7 +56,7 @@ string Strip0x(const string &hex)
 void HelpAndQuit(cxxopts::Options options)
 {
     cout << options.help({""}) << endl;
-    cout << "./ProofOfSpace generate" << endl;
+    cout << "./ProofOfSpace create" << endl;
```
justinengland, line 59: I assume this is related to a change in the underlying Proofofspace script? Any other impact here other than a rename of the function?

hoffmang9: This builds a c++ executable called ProofOfSpace and a python binary wheel. In chia-blockchain the app calls it `chia plots create` so when you switch to use the executable it's super easy to do this `./ProofOfSpace -k 32 create &`. Now that will work. Nothing depends on the ProofOfSpace command line arguments beyond itself.

src/cli.cpp
```diff
@@ -108,7 +108,7 @@ int main(int argc, char *argv[]) try {
 
     if (operation == "help") {
         HelpAndQuit(options);
-    } else if (operation == "generate") {
+    } else if (operation == "create") {
         cout << "Generating plot for k=" << static_cast<int>(k) << " filename=" << filename
```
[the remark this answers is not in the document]
hoffmang9: see earlier comment about this needed to mirror `chia plots create`

justinengland (approved):

PR #96, opened 2020-10-13
F1 optimization v2

rostislav:
Do not use Bits, provide F1 values in a simple array instead of a vector
of Bits. This increases performance of F1 calculation several times.

Changes in v2: do not use malloc, other small changes based on Arvid's feedback, rebase on master

### DIFF
--- a/src/calculate_bucket.hpp
+++ b/src/calculate_bucket.hpp
@@ -77,7 +77,9 @@ public:
     inline F1Calculator(uint8_t k, const uint8_t* orig_key)
     {
         uint8_t enc_key[32];
+        size_t buf_blocks = cdiv(k << kBatchSizes, kF1BlockSizeBits) + 1;
         this->k_ = k;
+        this->buf_ = new uint8_t[buf_blocks * kF1BlockSizeBits / 8];
 
         // First byte is 1, the index of this table
         enc_key[0] = 1;
@@ -87,7 +89,10 @@ public:
         chacha8_keysetup(&this->enc_ctx_, enc_key, 256, NULL);
     }
 
-    inline ~F1Calculator() {}
+    inline ~F1Calculator()
+    {
+        delete[] buf_;
+    }
 
     // Disable copying
     F1Calculator(const F1Calculator&) = delete;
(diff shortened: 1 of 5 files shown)


src/calculate_bucket.hpp
```diff
@@ -77,7 +77,9 @@ class F1Calculator {
     inline F1Calculator(uint8_t k, const uint8_t* orig_key)
     {
         uint8_t enc_key[32];
+        size_t buf_blocks = cdiv((k + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2;
         this->k_ = k;
+        this->buf_ = new uint8_t[buf_blocks * kF1BlockSizeBits / 8];
```
arvidn, line 82: `buf_` is only used in `CalculateBuckets()`, and it's overwritten every time that function is called. It's not obvious why it needs to be allocated in the constructor, why not allocate it temporarily in `CalculateBuckets()`?

Additionally, the size of this allocation is computed in a different way than the `n_blks` is computed. I imagine they end up being the same, since the test pass. It's not obvious that they are by reading the code though.

arvidn:
lgtm. Although it's hard to manually validate that all the calculations are correct. I take it we're confident in the unit test thoroughly covering validation of plots.

src/calculate_bucket.hpp
rostislav: The reason for allocating it here rather than in `CalculateBuckets` is that the latter function is "hot" and must run as fast as possible (it is the one being optimized in this commit).

Regarding the size of the allocation, there's a comment to `CalculateBuckets`, which states that `n` must not exceed `1 << kBatchSizes`. The allocation size is chosen so that it's sufficient for any `n` within the specified range.

arvidn: I see. so this is sort of like a "scratch buffer", that's reused to avoid the cost of heap allocation

src/calculate_bucket.hpp
```diff
@@ -154,69 +159,27 @@ class F1Calculator {
...
-            chacha8_get_keystream(&this->enc_ctx_, counter, 1, ciphertext_bytes);
-            Bits ciphertext(ciphertext_bytes, block_size_bits / 8, block_size_bits);
-            blocks.push_back(std::move(ciphertext));
-            ++counter;
+        uint64_t start = first_x * k_ / kF1BlockSizeBits;
+        // 'end' is one past the last keystream block number to be generated
+        // We first find the last used bit of the keystream, then convert it to
+        // block number and add one
+        uint64_t end = ((first_x + n) * k_ - 1) / kF1BlockSizeBits + 1;
+        uint64_t n_blks = end - start;
```
arvidn, line 171: I think it would make sense to have:

```
assert(n_blks <= cdiv((k_ + kExtraBits) << kBatchSizes, kF1BlockSizeBits) + 2);
```

rostislav: I did the math again and realized that this computation is wrong. I'll submit a new version with an assert here (a different one).

rostislav: Please see the new version with the assert. I'm reasonably sure that the math is right now and it's sufficient to ensure that `n <= (1U << kBatchSizes)`.

src/calculate_bucket.hpp
```diff
@@ -154,69 +159,29 @@ class F1Calculator {
...
-            block_size_bits;
-
-        std::vector<Bits> blocks;
-        uint64_t L = (counter * block_size_bits) / num_output_bits;
-        uint8_t ciphertext_bytes[kF1BlockSizeBits / 8];
+        uint64_t start = first_x * k_ / kF1BlockSizeBits;
+        // 'end' is one past the last keystream block number to be generated
+        // We first find the last used bit of the keystream, then convert it to
+        // block number and add one
+        uint64_t end = ((first_x + n) * k_ - 1) / kF1BlockSizeBits + 1;
```
arvidn, line 170: if the last bit is exactly divisible by `kF1BlockSizeBits`, this will be one too high, won't it?

would this be right?

```
uint64_t end = cdiv((first_x + n) * k_ - 1, kF1BlockSizeBits);
```

This would round it up to full blocks.

arvidn: actually, I think it should be:

```
uint64_t end = cdiv((first_x + n) * k_, kF1BlockSizeBits);
```

If this value ends up being too high, I suppose it's not the end of the world, since we just won't pull more than `n` bits from the buffer anyway. But the simpler expression we can use, the better I think.

rostislav: I did the math once more, and found that your second expression for `end` (without `- 1`) is mathematically equivalent to mine`. To put it another way, `cdiv(a, b)` can be defined as `(a + b - 1) / b` or `(a - 1) / b + 1`, assuming `a` is always positive.

I agree your version is a bit simpler, so I'll update the expression.

arvidn: you're right. in fact, would could defined `cdiv()` to use the `(a - 1) / b + 1` form if we believe it's better. It has less risk of overflow for instance, and maybe the arithmetic could be faster too, since it's just immediate values/constants.

rostislav: I've now pushed a new version with `cdiv` call. Regarding the implementation of `cdiv()` itself, I think the current one is better. The other version will not work with `a == 0`, and actually performs more operations than the current one when `b` is constant (3 vs 2 arithmetic ops).

The compilers also do a further optimization when `b` is a power of two, replacing division with right shift.

For these reasons I think it's preferable to keep the current version.

arvidn (approved):

PR #97, opened 2020-10-14
use unique_ptr for plotter_disk allocation

arvidn:
minor simplification of memory handling. The main change is to not zero the newly allocated memory, which saves a little bit of time


hoffmang9 (approved):
lgtm

PR #98, opened 2020-10-14
use emplace_back to construct FileDisk

arvidn:
`tablesize` was not used. I'm assuming that it's a left-over from some older version.


hoffmang9 (approved):
lgtm

PR #100, opened 2020-10-14
minor simplifications to DiskFile

arvidn:


src/disk.hpp
```diff
@@ -42,45 +41,42 @@ class Disk {
...
+    explicit FileDisk(const fs::path &filename)
     {
         filename_ = filename;
 
         // Opens the file for reading and writing
-        f_ = fopen(filename.c_str(), "w+b");
+        f_ = ::fopen(filename.c_str(), "w+b");
+        if (f_ == nullptr) {
+            throw InvalidValueException(
+                "Could not open " + filename.string() + ": " + ::strerror(errno));
```
arvidn, line 56: always report a failure to open a file (in the future, perhaps we should do something similar to when reading or writing fails, to wait 5 minutes and retry).

src/disk.hpp
```diff
@@ -42,45 +41,42 @@ class Disk {
...
 
     FileDisk(FileDisk &&fd)
     {
         filename_ = fd.filename_;
         f_ = fd.f_;
-        fd.f_ = NULL;
+        fd.f_ = nullptr;
     }
 
-    bool isOpen() { return (f_ != NULL); }
```
arvidn, line 65: is open was only used in one place, to check if the file opened successfully, and throw if it didn't. So I moved that directly into the constructor instead.

src/disk.hpp
```diff
@@ -139,22 +139,22 @@ class FileDisk : public Disk {
         } while (amtwritten != length);
     }
 
-    inline std::string GetFileName() override { return filename_.string(); }
+    std::string GetFileName() override { return filename_.string(); }
 
-    inline uint64_t GetWriteMax() const noexcept { return writeMax; }
+    uint64_t GetWriteMax() const noexcept { return writeMax; }
 
-    inline void Truncate(uint64_t new_size) override
```
arvidn, line 146: all these `inline` are redundant. The function bodies are already defined inline

hoffmang9 (approved):
Looks good to me

PR #104, opened 2020-10-19
Windows cmake build

arvidn:
These patches makes it possible to build the binaries on windows with cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,12 +41,30 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(pybind11-src)
 
+IF (MSVC)
+IF (CMAKE_BUILD_TYPE STREQUAL "RELEASE")
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /O2 /Oy")
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /O2 /Oy")
+ELSE()
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Ob1")
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Ob1")
+ENDIF()
+
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zi")
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Zi")
+
+ELSE()
 IF (CMAKE_BUILD_TYPE STREQUAL "RELEASE")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -41,6 +41,19 @@ FetchContent_Declare(
...
+IF (CMAKE_BUILD_TYPE STREQUAL "RELEASE")
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /O2 /Oy")
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /O2 /Oy")
+ELSE()
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Ob1")
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Ob1")
+ENDIF()
+
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zi")
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Zi")
```
arvidn, line 54: I'm not a cmake expert, but my understanding is that it doesn't abstract most compiler flags, such as these

uint128_t/uint128_t_config.include
```diff
@@ -1,13 +1,13 @@
 #ifndef _UINT128_T_CONFIG_
   #define _UINT128_T_CONFIG_
   #if defined(_MSC_VER)
-    #if defined(_DLL)
-      #define _UINT128_T_EXPORT __declspec(dllexport)
-      #define _UINT128_T_IMPORT __declspec(dllimport)
-    #else
+//    #if defined(_DLL)
```
arvidn, line 4: the problem is that `_DLL` appears to be defined whenever a shared library is being built, even if `uint128_t` *itself* isn't built as a shared library. The python module is a shared library. I believe this macro is defined by cmake, since it doesn't seem to be defined by setup tools.

A fix to this should probably be upstreamed to `uint128_t` too, since this kind of configuration only allows all-or-nothing static or shared library builds.

arvidn:
the LGTM issues come from `uint128_t` code, which previously wasn't analysed. there are two cases of dead code/redundant comparisons in `operator<<` and `operator>>`. As far as I can tell, they can't cause a bug.

arvidn:
I think the main immediate benefit of this PR is that the tests are built and run with MSVC on windows, which they aren't in `master` right now.

hoffmang9 (approved):
Now that I realize that this is a build method for Windows that is in addition to the cibuildwheel method and not replacing it (where the python tests are run on Windows too) this does look like a good idea.

Issue #103, opened 2020-10-17
Temp dirs not found plotting with b15 dev 47, to the root of a mount point on Windows


kiwihaitch:
On windows with a drive mountpoint at: C:\_\SM01\05 - plotting to this directory path fails with a Temp dir not found. If I create a subdir below it, eg C:\_\SM01\05\PLOT - plot works properly.

Log file attached.

hoffmang9:
Additional notes:

1. That path as tempdir works under Beta 15
2. It's pretty clearly failing in C++ and not the python layer.

hoffmang9:
Windows Server 2016

kiwihaitch:
OS: Server 2016. Log re-attached 

[b15.47-plot-error.txt](https://github.com/Chia-Network/chiapos/files/5395066/b15.47-plot-error.txt)

arvidn:
I tried to reproduce this with the C++ plotting tool, on windows. No luck so far, I'm running:

```
Debug\ProofOfSpace.exe create -k 32 -b 4528 -u 128 -s 65536 -t C:\_\SM01\05 -2 C:\_\SM01\05 -d C:\_\SM01\05
```

(and make sure I create the directory `c:\_\SM01\05`)

Looking at the code, this error comes from:

```
if (!fs::exists(tmp_dirname)) {
   throw InvalidValueException("Temp directory " + tmp_dirname + " does not exist");
}
```

Since the error message says:

```
Caught plotting error: Temp directory C:\_\SM01\05 does not exist
```

It really suggests that the directory "C:\_\SM01\05" indeed does not exist. It's possible that there could be other conditions causing `fs::exists()` to return false though.

kiwihaitch:
The Path C:\_\SM01\05 has to be an NTFS Mountpoint for it to fail (ie has another drive mounted there rather than have it's own drive letter). If a directory below the mountpoint is used eg: C:\_\SM01\05\Plot then the plotter is fine. It's only in the specific case of attemptting to plot to the root of the mountpoint.

![image](https://user-images.githubusercontent.com/14339794/96458238-4ef8a380-11e6-11eb-9806-b9556e3ff1b6.png)

arvidn:
oh, I see. The log said "C:\_\..." so I missed that you're using the per-drive working directory.

arvidn:
I have reproduced the issue

arvidn:
https://github.com/gulrak/filesystem/issues/75

arvidn:
[...]

PR #105, opened 2020-10-20
hack filesystem to work around issue in parsing mount-points

arvidn:
This will fix the specific issue reported here: https://github.com/Chia-Network/chiapos/issues/103
However, I have not reviewed all our use of the filesystem to ensure this won't break anything else.
It's likely to break *some* uses of the filesystem library, I think it's mostly exotic use though.

I only run windows in a VM, so I have not even smoke tested this patch. Someone on windows should do that before we land it.


hoffmang9 (approved):
Looks like a perfect dirty hack that needs testing to me!

PR #106, opened 2020-10-22
Improve performance of SliceInt64FromBytes

aqk:
This attempts to preserve the semantics of the recent out-of-bounds bugfix.

### DIFF
--- a/src/util.hpp
+++ b/src/util.hpp
@@ -237,8 +237,19 @@ namespace Util {
         }
 
         uint64_t tmp = 0;
-        memcpy(&tmp, bytes, std::min(uint32_t(sizeof(uint64_t)), cdiv(start_bit + num_bits, 8)));
-        tmp = bswap_64(tmp);
+        auto const num_bytes = cdiv(start_bit + num_bits, 8);
+        switch(num_bytes) {
+        default: tmp |= uint64_t(bytes[7]);
+        case 7: tmp |= uint64_t(bytes[6]) << (1*8);
+        case 6: tmp |= uint64_t(bytes[5]) << (2*8);
+        case 5: tmp |= uint64_t(bytes[4]) << (3*8);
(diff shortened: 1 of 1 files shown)


hoffmang9:
Kicking off a reference run of a k=30. Time to beat is 12725.418 seconds.

Time with only the bug fix reverted instead and with the other improvements is 11458.487 seconds.

arvidn:
Adam, did you look at the code this produces compares to the existing code?

Another thing to try is to spell out the condition as an if-statement (instead of min()); in the hopes the branch predictor can kick in. But maybe the problem is that this uses an additional slot in the branch predictor table.

arvidn:
this gives me an idea. The only reason (as far as I can tell) that we need the call to `std::min()` currently, is because `SliceInt64FromBytesFull()` wants to use this function in a funky way, where `start_bit + num_bits > 64`, which causes us to truncate. `SliceInt64FromBytesFull()` will then fixup the last bits from the 9th byte.

If `SliceIn64FromBytesFull()` instead used this "duff's device"-like implementation, I think `SliceInt64FromBytes()` could be simplified.

The main cost, though, is probably using a variable size in `memcpy()` compared to always copying 8 bytes, as we did previously. With fixed 8 bytes, it probably turns into a single `mov`.

arvidn (approved):
If this turns out to be faster, I think it's good. I suspect that the main problem is that the size parameter to `memcpy()` isn't a constant anymore though.

hoffmang9:
11780.493 seconds.

hoffmang9:
Beta 15 time for reference: Total time = 11670.208 seconds.

arvidn:
I think this patch has been superseded now

hoffmang9:
@aqk - should I just delete the branch?

hoffmang9:
@aqk Should I delete the branch too?

PR #108, opened 2020-10-23
Add extra 7 bytes to buffers passed to Util::SliceInt64FromBytes

hoffmang9:


wjblanke (approved):
Looks good to me. Also tested with k27 and k29.

PR #109, opened 2020-10-23
f1 parallel

wjblanke:


hoffmang9 (approved):
A-OK - squeezing out those cycles!

PR #110, opened 2020-10-27
use uniqe_ptr<> instead of raw pointers to hold arrays

arvidn:
simplifies memory management. Remove unused allocation in phase3


hoffmang9 (approved):
Looks good

PR #112, opened 2020-10-29
Sort manager simplification

arvidn:
This change is split up into mostly independent commits, it might be easier to review each commit rather than the whole change.

The main change is to join all the vectors of members used for each bucket, into a single vector of a struct with all the per-bucket members (`bucket_t`).

The convention in the rest of the code seems to use a `_` suffix on members, rather than always accessing them through `this`, so I'm transitioning towards such naming in `SortManager` too (and using `this` is more of a Java thing and not very common in C++).

`inline` are redundant on member functions that are defined inline (as it's implied).

I transitioned two memory allocations to use `unique_ptr<>` instead of raw `new` and `delete[]`.

Instead of assigning values to members in the constructor, I introduced an initializer list to initialize the members instead, as well as added `const` in a few places to ease reading.

### DIFF
--- a/src/sort_manager.hpp
+++ b/src/sort_manager.hpp
@@ -66,17 +66,17 @@ public:
         std::vector<fs::path> bucket_filenames = std::vector<fs::path>();
 
         for (size_t bucket_i = 0; bucket_i < num_buckets; bucket_i++) {
-            this->mem_bucket_pointers.push_back(memory + bucket_i * size_per_bucket);
-            this->mem_bucket_sizes.push_back(0);
-            this->bucket_write_pointers.push_back(0);
             std::ostringstream bucket_number_padded;
             bucket_number_padded << std::internal << std::setfill('0') << std::setw(3) << bucket_i;
 
-            fs::path bucket_filename =
+            fs::path const bucket_filename =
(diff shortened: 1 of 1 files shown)


mariano54:
Hey arvid, these changes seem good. I am prioritizing on the consensus so I have not looked at these changes in detail, and will not be working on chiapos for a while, but feel free to merge whatever changes you think are good.

src/sort_manager.hpp
```diff
@@ -224,44 +235,48 @@ class SortManager {
...
+        bucket_t& b = buckets_[bucket_i];
+
+        uint64_t const start_write = b.write_pointer;
+        uint64_t const write_len = b.size * this->entry_size;
 
         // Flush the bucket to disk
-        bucket_files[bucket_i].Write(start_write, this->mem_bucket_pointers[bucket_i], write_len);
-        this->bucket_write_pointers[bucket_i] += write_len;
+        b.file.Write(start_write, b.mem, write_len);
+        b.write_pointer += write_len;
```
aqk, line 239: This array-of-struct notation is easier to read.

aqk (approved):
Read & Approved.

PR #116, opened 2020-11-03
more use of std::unique_ptr in phase RunPhase1, using local num_threads

aldenml:


src/phase1.hpp
```diff
@@ -744,19 +741,19 @@ std::vector<uint64_t> RunPhase1(
 
         Timer computation_pass_timer;
 
-        THREADDATA* td = new THREADDATA[num_threads];
-        auto* mutex = new Sem::type[num_threads];
+        auto td = std::make_unique<THREADDATA[]>(num_threads);
+        auto mutex = std::make_unique<Sem::type[]>(num_threads);
```
arvidn, line 745: this is correct. One thing to keep in mind for similar transformations is that this will initialize all the elements, whereas the previous version didn't. C++20 introduced `make_unique_for_overwrite` for this purpose, but I don't think we want to require C++20.

arvidn (approved):

src/phase1.hpp
aldenml: totally agree with the comment, maybe `make_unique_for_overwrite` can be added as a utility for the time being.

PR #117, opened 2020-11-03
sort_manager "quicksort" parameter

arvidn:
This change makes the sort strategy (uniform sort, quicksort, or quicksort last bucket) a property of the sort manager, rather than a parameter to `ReadEntry()` et.al.

Partly this is a simplification, but it also prepares the sort manager to implement the Disk interface, where `Read()` doesn't want to have to care about which sorting should be used.

It's split up into two commits, as it was developed in two steps:

1. introduce the constructor parameter and member to store the sorting strategy. Assert that whatever argument is passed to `ReadEntry()` matches the member, and all tests pass.
2. Remove the member.

This is intended to not introduce any functional differences, just how and where in the code the sorting is determined.

### DIFF
--- a/src/phase1.hpp
+++ b/src/phase1.hpp
@@ -226,7 +226,7 @@ void* phase1_thread(THREADDATA* ptd)
             if (!first_thread) {
                 Sem::Wait(ptd->theirs);
             }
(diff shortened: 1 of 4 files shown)


hoffmang9 (approved):
Looks good

PR #118, opened 2020-11-04
use unique_ptr instead of malloc in phase1 F1thread

aldenml:


src/phase1.hpp
```diff
@@ -553,7 +553,7 @@ void* F1thread(THREADF1DATA* ptd)
 
     uint64_t right_buf_entries = 1 << (kBatchSizes);
 
-    uint64_t* f1_entries = (uint64_t*)malloc((1 << kBatchSizes) * sizeof(*f1_entries));
```
aldenml, line 556: @arvidn I'm a little puzzled by this, the use of `malloc` seems so specific, that maybe there is some non-obvious reason for it. Also, isn't a problem dereferencing `f1_entries` at this moment? and if that's the case, is there a way for the sanitizer to capture that?

arvidn: I don't think there's a reason to use `malloc()` over `new` here. It was probably just out of habit. One reason I can think of (that doesn't apply here) would be if you want an allocation that's suitably aligned for any type. But we know the type here, so 8-byte alignment is sufficient.

I considered making this change, and I think it's an improvement, but I also think that this allocation probably is unnecessary all together. The work done in `CalculateBuckets()`, that fills in this array, could be interleaved with the loop, which would render the buffer unnecessary. There's just one write pass and one read pass over the buffer. (for example, the loop-body could be passed as a function object to `CalculateBuckets()`, to pass the results on in small portions).

That would probably be an improvement as well, however, this part of the code is already relatively fast compared to many other places, so it's not very high up on my list.

arvidn (approved):

PR #119, opened 2020-11-04
increase fidelity in disk I/O logging

arvidn:
to include reads and writes per file. This logging is still only enabled as a compile-time option (by modifying `src/disk.hpp`).


hoffmang9 (approved):
lgtm

PR #122, opened 2020-11-06
@jaraco @hoffmang9 Don't pass empty string to msvc linker. Ref pypa/setuptools#2417

hoffmang9:
Also removed SETUPTOOLS_USE_DISTUTILS=stdlib from the Windows environment


wjblanke (approved):
Gah!

PR #123, opened 2020-11-09
check that plotting doesn't use too much RAM on CI (plot -k 27)

arvidn:


hoffmang9 (approved):
"ProofOfSpace used 901820 kiB of memory"

arvidn:
> "ProofOfSpace used 901820 kiB of memory"

the k=27 plot in CI used 2504960 kiB. that's just 900 MiB, you didn't see that in CI, did you?

hoffmang9:
That was what I saw on the push here - https://github.com/Chia-Network/chiapos/runs/1376084897?check_suite_focus=true#step:5:464

arvidn:
interesting. I just double checked that the `M` format specifier in `time` indeed means:

> Maximum resident set size of the process during its lifetime, in Kilobytes.

The `on: push` and `on: pull_request` builds are quite different. Perhaps `time` isn's as reliable as I though.

https://github.com/Chia-Network/chiapos/pull/123/checks?check_run_id=1376084713#step:5:456
https://github.com/Chia-Network/chiapos/pull/123/checks?check_run_id=1376084897#step:5:464

hoffmang9:
I assume there is some potential variability to the Azure underlying VMs but I wouldn't think that would change our memory footprint...

PR #124, opened 2020-11-09
simplify parallel F1 calculation

arvidn:
ordering of entries begin put into the sort manager is not important, since it will be sorted anyway. So this uses a single mutex instead of cascading semaphores.

### DIFF
--- a/src/phase1.hpp
+++ b/src/phase1.hpp
@@ -34,6 +34,7 @@
 #include <vector>
 #include <thread>
 #include <memory>
+#include <mutex>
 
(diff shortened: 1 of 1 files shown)


hoffmang9 (approved):
lgtm

arvidn:
this patch *is* making one important assumption (which I clearly think is a safe one); the order we put entries into the SortManager is not important.

The case where it *could* be important would be if we have entries that are *equal* and whose ordering is arbitrary. Previously such entries would be ordered consistently, and they aren't anymore. I don't get the impression we have such entries though.

PR #125, opened 2020-11-09
Nogulrakwjb

wjblanke:
Well, Gulrak for Mac only

### DIFF
--- a/src/plotter_disk.hpp
+++ b/src/plotter_disk.hpp
@@ -42,7 +42,6 @@
 namespace fs = ghc::filesystem;
 #else
 #include <filesystem>
-namespace fs = std::filesystem;
 #endif
 
 #include "calculate_bucket.hpp"
--- a/src/plotter_disk.hpp
+++ b/src/plotter_disk.hpp
@@ -41,7 +41,13 @@
 #include "../lib/include/filesystem.hpp"
 namespace fs = ghc::filesystem;
 #else
+#ifdef __cpp_lib_filesystem
 #include <filesystem>
+#elif __cpp_lib_experimental_filesystem
+#include <experimental/filesystem>
+#else
+#error "no filesystem support ='("
+#endif
 #endif
 
 #include "calculate_bucket.hpp"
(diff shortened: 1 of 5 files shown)


src/plotter_disk.hpp
```diff
@@ -36,8 +36,18 @@
...
+#ifdef __APPLE__
+// std::filesystem is not supported on Mojave
 #include "../lib/include/filesystem.hpp"
 namespace fs = ghc::filesystem;
+#else
+#if __has_include(<filesystem>)
+#include <filesystem>
+#else
+#include <experimental/filesystem>
+#endif
```
arvidn, line 48: you still need to alias the namespace, don't you?

For example:

```
namespace fs = std::experimental::filesystem;
```

src/plotter_disk.hpp
```diff
@@ -36,8 +36,18 @@
 #define _HAS_STD_BYTE 0
 #define NOMINMAX
 
+#ifdef __APPLE__
+// std::filesystem is not supported on Mojave
 #include "../lib/include/filesystem.hpp"
 namespace fs = ghc::filesystem;
+#else
+#if __has_include(<filesystem>)
```
arvidn, line 44: if you really want to produce nice errors, you can check for the existence of this feature by:

```
#ifdef __has_include
#if __has_include(<...>)
#endif
#endif
```

src/sort_manager.hpp
```diff
@@ -25,9 +25,17 @@
...
 #include "../lib/include/filesystem.hpp"
-
 namespace fs = ghc::filesystem;
+#else
+#if __has_include(<filesystem>)
+#include <filesystem>
+#else
+#include <experimental/filesystem>
+#endif
+#endif
```
arvidn, line 38: it might be worth putting this in a new header "filesystem.hpp", or something like that

src/plotter_disk.hpp
wjblanke: seems like filesystem already defines it

src/sort_manager.hpp
wjblanke: moved to chia_filesystem.hpp

src/plotter_disk.hpp
wjblanke: added to chia_filesystem.hpp. it tries <filesystem> if __has_include is missing.

hoffmang9 (approved):
Shippin' it.

PR #126, opened 2020-11-10
some phase1 modernization

arvidn:
make more variables immutable, use structured binding instead of std::get<>


hoffmang9 (approved):
Looks good

PR #127, opened 2020-11-10
Modernize sort

arvidn:
replaces `class` with `namespace` (where classes are used as namespaces). Makes variables immutable. use `unique_ptr` for memory allocations. use constant instead of macro.


hoffmang9 (approved):
Approving and merging while ignoring that odd problem with GH Actions.

PR #129, opened 2020-11-11
Windowstimewjb

wjblanke:
correct windows cpu %

### DIFF
--- a/src/util.hpp
+++ b/src/util.hpp
@@ -73,8 +73,9 @@ std::ostream &operator<<(std::ostream &strm, uint128_t const &v)
 
 class Timer {
 public:
-    Timer() : wall_clock_time_start_(std::chrono::steady_clock::now())
+    Timer()
     {
+        wall_clock_time_start_ = std::chrono::steady_clock::now();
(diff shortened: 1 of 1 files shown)


src/util.hpp
```diff
@@ -35,6 +35,10 @@ template <typename Int>
 constexpr inline Int cdiv(Int a, int b) { return (a + b - 1) / b; }
 
 #ifdef _WIN32
+#define _HAS_STD_BYTE 0
+#define NOMINMAX
+#include "windows.h"
+#include "processthreadsapi.h"
```
arvidn, line 41: if I may be a bit pedantic; `windows.h` and `processthreadsapi.h` should probably be included with `<` and `>`, since they are system headers.
Also, I don't think anything is looking at `_HAS_STD_BYTE`. It has a reserved name, so if anything, I would expect it to be defined by the implementation.

arvidn (approved):

src/util.hpp
wjblanke: fixed

PR #132, opened 2020-12-02
changed scope of match_indexes to make sure they are always reset eve…

wjblanke:
…n if R bucket is empty https://github.com/Chia-Network/chia-blockchain/issues/504


arvidn (approved):
it would probably be (marginally) faster to call `clear()` instead, to allow reusing the allocation

PR #134, opened 2020-12-05
using range for loop and removed unnecessary returns in phase1.hpp

aldenml:


aldenml:
@arvidn gentle request for review here :)

src/phase1.hpp
```diff
@@ -520,7 +517,7 @@ void* phase1_thread(THREADDATA* ptd)
         Sem::Post(ptd->mine);
     }
 
-    return 0;
```
arvidn, line 523: as far as I can tell, this is the only return statement in this function. This function doesn't *have* to return anything. I think it would be better to define the function to not return anything and just remove the return statement

src/phase1.hpp
```diff
@@ -570,7 +567,7 @@ void* F1thread(int const index, uint8_t const k, const uint8_t* id, std::mutex*
         }
     }
 
-    return 0;
```
arvidn, line 573: same thing goes for this function

arvidn (approved):
looks good apart from the two comments

aldenml:
done and squashed

wjblanke:
I'd like to merge this, but timings for this branch are slower than the base branch. I think that is because of differences between how iterator versus [] is implemented. It may be Ubuntu specific, as different platforms implement STL differently. Let me know if you see different timings for large k.

aldenml:
Hi @wjblanke, I'm still waiting for HW to play with a bigger `k`, but I just noticed that this branch is behind `master` 17 commits, I will rebase to be on top. Do you mind test again? (this type of change normally is a net-win in relase mode)

arvidn:
> I think that is because of differences between how iterator versus [] is implemented.

an iterator is a pointer, so it's not really a matter of standard library implementation. It's possible that the alias analysis is having a harder time knowing that the entries the iterator points to don't alias anything else, whereas using an index, the fact that the element belongs to the array is obvious at each access.

I would also think that allocating `not_dropped` dynamically could be costly. If we have a sense of how large that array is up-front, we could probably save a fair amount of time by `reserve()`ing it.

arvidn:
another likely performance boost would be to split out the `used` field and stick that in a separate bitfield, but still associated with the corresponding entry in the `bucket_L` vector. At least of that's `false` a lot, it would save us the cache miss of pulling in the element from `bucket_L`. These comments aren't suggestions to extend *this* PR though.

But, splitting out the redundant return value from the thread function into a separate PR might get that landed at least. That seems like a no-brainer.

aldenml:
not really looking for a contentious merge, the suggestions are good and I totally understand the need to be obsessive on not having performance regressions in this project. I will try to set up my env to back any future PR with numbers.

PR #135, opened 2020-12-07
clear input files for tables

arvidn:
once they have been transformed into the that will be used in the next phase. This saves disk space.

My measurements of peak disk usage are (for a k24 plot):

```
current master: 1602MB
this patch: 964MB
prior to bitfield-patch: 1054MB
```

(these aren't exact, it's based on polling `ls -la` every 2 seconds during the plot)

There's still an issue that we leave *empty* files lingering longer than it seems necessary, but it's mostly a cosmetic issue.

### DIFF
--- a/src/phase2.hpp
+++ b/src/phase2.hpp
@@ -235,6 +235,16 @@ Phase2Results RunPhase2(
         }
         current_bitfield.swap(next_bitfield);
         next_bitfield.clear();
+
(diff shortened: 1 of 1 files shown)


hoffmang9 (approved):
now using slightly less temp space than beta 17. 20-20% slower overall - 4% more memory

PR #136, opened 2020-12-07
reset read/write cursor on file close

arvidn:
when closing a file, the file's read and write cursor no longer exists, reset our prediction of it (this affects whether we call seek() or not before reading and writing)


hoffmang9 (approved):
looks good

PR #140, opened 2020-12-09
Use try to avoid race condition on file glob

hoffmang9:
When the background polling thread is running it can have a .tmp file found in it's glob of the directory that gets deleted before it can stat it.


wjblanke (approved):
AOK!

PR #141, opened 2020-12-10
Nostlwjb

wjblanke:
use stack for no overhead. improves b17 by 11%

### DIFF
--- a/src/calculate_bucket.hpp
+++ b/src/calculate_bucket.hpp
@@ -276,11 +276,14 @@ public:
     // bucket length, we can store all the R values and lookup each of our 32 candidates to see if
     // any R value matches. This function can be further optimized by removing the inner loop, and
     // being more careful with memory allocation.
-    inline std::vector<std::pair<uint16_t, uint16_t>> FindMatches(
+    inline void FindMatches(
         const std::vector<PlotEntry>& bucket_L,
-        const std::vector<PlotEntry>& bucket_R)
+        const std::vector<PlotEntry>& bucket_R,
+        uint16_t *idx_L,
+        uint16_t *idx_R,
+        uint32_t &idx_count)
     {
-        std::vector<std::pair<uint16_t, uint16_t>> matches;
+        //std::vector<std::pair<uint16_t, uint16_t>> matches;
         uint16_t parity = (bucket_L[0].y / kBC) % 2;
 
         for (size_t i = 0; i < rmap_clean.size(); i++) {
@@ -306,12 +309,14 @@ public:
             for (uint8_t i = 0; i < kExtraBitsPow; i++) {
                 uint16_t r_target = L_targets[parity][r][i];
                 for (size_t j = 0; j < rmap[r_target].count; j++) {
-                    matches.push_back(std::make_pair(pos_L, rmap[r_target].pos + j));
+                    if(idx_L != NULL) {
+                        idx_L[idx_count]=pos_L;
+                        idx_R[idx_count]=rmap[r_target].pos + j;
+                    }
+                    idx_count++;
                 }
             }
         }
-
-        return matches;
     }
 
 private:
--- a/src/calculate_bucket.hpp
+++ b/src/calculate_bucket.hpp
@@ -276,13 +276,13 @@ public:
     // bucket length, we can store all the R values and lookup each of our 32 candidates to see if
     // any R value matches. This function can be further optimized by removing the inner loop, and
     // being more careful with memory allocation.
-    inline uint32_t FindMatches(
+    inline int32_t FindMatches(
         const std::vector<PlotEntry>& bucket_L,
         const std::vector<PlotEntry>& bucket_R,
         uint16_t *idx_L,
         uint16_t *idx_R)
     {
-        uint32_t idx_count = 0;
+        int32_t idx_count = 0;
         uint16_t parity = (bucket_L[0].y / kBC) % 2;
 
         for (size_t i = 0; i < rmap_clean.size(); i++) {
--- a/src/calculate_bucket.hpp
+++ b/src/calculate_bucket.hpp
@@ -276,14 +276,13 @@ public:
     // bucket length, we can store all the R values and lookup each of our 32 candidates to see if
     // any R value matches. This function can be further optimized by removing the inner loop, and
     // being more careful with memory allocation.
-    inline void FindMatches(
+    inline uint32_t FindMatches(
         const std::vector<PlotEntry>& bucket_L,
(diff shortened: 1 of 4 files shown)


hoffmang9 (approved):
Looks good but will kick off a k32

src/calculate_bucket.hpp
```diff
@@ -276,11 +276,14 @@ class FxCalculator {
...
     // any R value matches. This function can be further optimized by removing the inner loop, and
     // being more careful with memory allocation.
-    inline std::vector<std::pair<uint16_t, uint16_t>> FindMatches(
+    inline void FindMatches(
         const std::vector<PlotEntry>& bucket_L,
-        const std::vector<PlotEntry>& bucket_R)
+        const std::vector<PlotEntry>& bucket_R,
+        uint16_t *idx_L,
+        uint16_t *idx_R,
+        uint32_t &idx_count)
```
arvidn, line 284: if we want to squeeze the last bit of throughput out of this, it may be better to take this by value and return the new index. That way the counter can live in a register and not have to spill to RAM

src/calculate_bucket.hpp
```diff
@@ -306,12 +309,14 @@ class FxCalculator {
             for (uint8_t i = 0; i < kExtraBitsPow; i++) {
                 uint16_t r_target = L_targets[parity][r][i];
                 for (size_t j = 0; j < rmap[r_target].count; j++) {
-                    matches.push_back(std::make_pair(pos_L, rmap[r_target].pos + j));
+                    if(idx_L != NULL) {
```
arvidn, line 312: can `idx_L` really be null here? if it can, what's the expected behavior, to *just* increment the counter? This check probably won't be expensive here since it will make the same branch every time through the loop. But it looks a bit suspicious to support it being null to begin with. It could be tested once and have two separate loops for the two behaviors, to hoist the check out of the inner loop

src/phase1.hpp
```diff
@@ -266,20 +266,23 @@ void* phase1_thread(THREADDATA* ptd)
...
+                uint32_t idx_count=0;
 
                 if (!bucket_L.empty()) {
                     not_dropped.clear();
 
                     if (!bucket_R.empty()) {
                         // Compute all matches between the two buckets and save indeces.
-                        match_indexes = f.FindMatches(bucket_L, bucket_R);
-
+                        f.FindMatches(bucket_L, bucket_R, idx_L, idx_R, idx_count);
```
arvidn, line 279: My bet would be that this is by far the greatest performance improvement in this patch. This is something I've been eyeing for a long time, but it wasn't obvious to me how `match_indices` was used and whether anything relied on it just being the most recent batch of matches. I suppose this is clearer now as you narrowed the scope of the `match_indexes` container

src/phase1.hpp
```diff
@@ -266,20 +266,23 @@ void* phase1_thread(THREADDATA* ptd)
...
+//std::cout << "matches " << idx_count << std::endl;
```
arvidn, line 280: you should probably remove this before landing the patch

src/phase1.hpp
```diff
@@ -266,20 +266,23 @@ void* phase1_thread(THREADDATA* ptd)
                 // so now we can compare entries in both buckets to find matches. If two entries
                 // match, match, the result is written to the right table. However the writing
                 // happens in the next iteration of the loop, since we need to remap positions.
-                std::vector<std::pair<uint16_t, uint16_t>> match_indexes;
+                // std::vector<std::pair<uint16_t, uint16_t>> match_indexes;
+                uint16_t idx_L[10000];
+                uint16_t idx_R[10000];
```
arvidn, line 271: I would also be a bit worried about the fixed size of this array. There isn't a way to reliably compute an upper bound of this? At the very least, I think `idx_count` should be checked after the call to `FindMatches()` to ensure its < `std::size(idx_R)` and abort if it isn't. We would have invoked undefined behavior already, but at least we would have a chance to abort

src/phase1.hpp
```diff
@@ -266,20 +266,23 @@ void* phase1_thread(THREADDATA* ptd)
...
                     if (!bucket_R.empty()) {
                         // Compute all matches between the two buckets and save indeces.
-                        match_indexes = f.FindMatches(bucket_L, bucket_R);
-
+                        f.FindMatches(bucket_L, bucket_R, idx_L, idx_R, idx_count);
+//std::cout << "matches " << idx_count << std::endl;
                         // We mark entries as used if they took part in a match.
-                        for (auto [idx_L, idx_R] : match_indexes) {
-                            bucket_L[idx_L].used = true;
+                        for (uint32_t i=0; i < idx_count; i++) {
```
arvidn, line 282: the counter should be either signed or a full 64 bits here. Otherwise there's a risk the compiler think you are relying on modulo-2 arithmetic.

src/phase1.hpp
```diff
@@ -350,9 +353,9 @@ void* phase1_thread(THREADDATA* ptd)
                     current_entries_to_write = std::move(future_entries_to_write);
                     future_entries_to_write.clear();
 
-                    for (auto [idx_L, idx_R] : match_indexes) {
-                        PlotEntry& L_entry = bucket_L[idx_L];
-                        PlotEntry& R_entry = bucket_R[idx_R];
+                    for (uint32_t i=0; i < idx_count; i++) {
```
arvidn, line 356: ditto

src/calculate_bucket.hpp
wjblanke: Yes, in tests we only need the count

src/calculate_bucket.hpp
wjblanke: done

src/phase1.hpp
wjblanke: removed

src/phase1.hpp
wjblanke: checked

src/phase1.hpp
wjblanke: signed. old habits die hard!

src/phase1.hpp
wjblanke: yes

hoffmang9:
Wut?
```
Total time for ProofOfSpace: 1353.2616691589355 seconds
Total temp space used: 8.785296318121254 GiB
Memory used: 5157.76171875 MiB
Total plot size: 2.6943268217146397 GiB
```

hoffmang9:
NM - I was confused on temp space as the bug I'm about to fix was messing with the temp space calculations.

PR #142, opened 2020-12-10
Have ProofOfSpace mirror chia-blockchain's -r - fix temp space accumulator

hoffmang9:
Using two threads is a nearly 30% speedup and we should move to that once we've gotten a new baseline so that the CI runs faster.


wjblanke (approved):
AOK

PR #146, opened 2020-12-12
Resources args

hoffmang9:
- Add arguments to plot-resources.py monitor
- Also plot a k=27 with bitfield disabled.


wjblanke (approved):
AOK

PR #150, opened 2021-01-04
Update to 3584MiB RAM buffer -num_threads 2

hoffmang9:
Once upstreamed to chia-blockchain - should revert src/cmds/plots.py to use 0's instead of defaults and rely on these


wjblanke (approved):
aok

PR #154, opened 2021-01-19
Fix buffered disk

arvidn:
This fix two performance issues in one layer of the disk abstraction `BufferedDisk`.
It also turns an assert into a warning log. This was uncovered by me hitting this assert while testing a k32 under address sanitizer.

descriptions are inline.

### DIFF
--- a/src/disk.hpp
+++ b/src/disk.hpp
@@ -232,14 +232,23 @@ struct BufferedDisk : Disk
         NeedReadCache();
         // all allocations need 7 bytes head-room, since
         // SliceInt64FromBytes() may overrun by 7 bytes
-        if (read_buffer_start_ <= begin && read_buffer_start_ + read_buffer_size_ > begin + length + 7) {
+        if (read_buffer_start_ <= begin
+            && read_buffer_start_ + read_buffer_size_ >= begin + length
+            && read_buffer_start_ + read_ahead >= begin + length + 7)
+        {
             // if the read is entirely inside the buffer, just return it
             return read_buffer_.get() + (begin - read_buffer_start_);
         }
-        else if (begin >= read_buffer_start_ || begin == 0) {
+        else if (begin >= read_buffer_start_ || begin == 0 || read_buffer_start_ == std::uint64_t(-1)) {
+
             // if the read is beyond the current buffer (i.e.
             // forward-sequential) move the buffer forward and read the next
-            // buffer-capacity number of bytes
+            // buffer-capacity number of bytes.
+            // this is also the case we enter the first time we perform a read,
+            // where we haven't read anything into the buffer yet. Note that
+            // begin == 0 won't reliably detect that case, sinec we may have
+            // discarded the first entry and start at some low offset but still
+            // greater than 0
             read_buffer_start_ = begin;
             uint64_t const amount_to_read = std::min(file_size_ - read_buffer_start_, read_ahead);
             disk_->Read(begin, read_buffer_.get(), amount_to_read);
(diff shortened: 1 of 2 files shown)


src/disk.hpp
```diff
@@ -232,14 +232,23 @@ struct BufferedDisk : Disk
         NeedReadCache();
         // all allocations need 7 bytes head-room, since
         // SliceInt64FromBytes() may overrun by 7 bytes
-        if (read_buffer_start_ <= begin && read_buffer_start_ + read_buffer_size_ > begin + length + 7) {
+        if (read_buffer_start_ <= begin
+            && read_buffer_start_ + read_buffer_size_ >= begin + length
+            && read_buffer_start_ + read_ahead >= begin + length + 7)
```
arvidn, line 237: This condition needs to check two separate criteria:

1. The read request is within the cache. i.e. we have all these bytes in the cache
2. The read request has at least 7 bytes of margin before the end of the *allocation*

The existing check conflates these two and makes sure there are at least 7 bytes of *valid* cache after the current request. This causes the last few requests to hit the disk and wipe the cache even though they don't need to.

src/disk.hpp
```diff
@@ -232,14 +232,23 @@ struct BufferedDisk : Disk
...
+        {
             // if the read is entirely inside the buffer, just return it
             return read_buffer_.get() + (begin - read_buffer_start_);
         }
-        else if (begin >= read_buffer_start_ || begin == 0) {
+        else if (begin >= read_buffer_start_ || begin == 0 || read_buffer_start_ == std::uint64_t(-1)) {
```
arvidn, line 242: the `begin == 0` condition here was meant to catch the initial read, since we perform forward sequential scans. However, with the `FilteredDisk` (where we save a read/write pass to compact the table) our first read may not actually be at 0, but a few entries in (if the first ones were dropped).

This fixes the condition to also say, "if we don't have a cache yet" (i.e. `read_buffer_start == std::uint64_t(-1)`)

src/disk.hpp
```diff
@@ -248,7 +257,13 @@ struct BufferedDisk : Disk
...
         else {
             // ideally this won't happen
-            assert(false);
+            std::cout << "Disk read position regressed. It's optimized for forward scans. Performance may suffer\n"
+                << "   read-offset: " << begin
+                << " read-length: " << length
+                << " file-size: " << file_size_
+                << " read-buffer: [" << read_buffer_start_ << ", " << read_buffer_size_ << "]"
+                << " file: " << disk_->GetFileName()
+                << '\n';
```
arvidn, line 266: This is still not expected to happen, so if it does, it would be good to know

tests/test.cpp
```diff
@@ -900,7 +900,124 @@ void test_bitfield_size(int const size)
...
+        // print a warning
+        for (int i = num_test_entries / 2 - 1; i > num_test_entries / 2 + 200; --i) {
+            auto const val = *reinterpret_cast<std::uint32_t const*>(fd.Read(i * 4, 4));
+            CHECK((i * 2) == val);
+        }
+    }
+/*
+    SECTION("empty bitfield")
+    {
+        BufferedDisk bd(&d, num_test_entries * 4);
```
arvidn, line 1017: This commented-out test is a reminder that the `FilteredDisk` requires that at least 1 entry is not filtered. i.e. it cannot model a file where every entry is filtered. It doesn't seem worthwhile plugging this hole, since it won't happen in this program

wjblanke (approved):
AOK

src/disk.hpp
arvidn: if there truly is a performance regression with this patch (which seems unlikely to me), I would think it's the fact that there's one more condition here.

PR #155, opened 2021-01-19
Handle SIGINT (exit and cleanup tmp files for current plot)

kd637xx:
`CTRL+C` or `kill -2 <pid>` now gracefully ends plotting (cleans current plot temp files) - in CLI

### DIFF
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -88,9 +88,9 @@ void cleanup(const string& tmpdir)
 
 void SigHandler(int s)
 {
-    cout << endl << "Exiting.." << endl;        
+    cout << endl << "Exiting .." << endl;        
     cleanup(tempdir);
-    cleanup(tempdir2);
+    if (tempdir.compare(tempdir2) != 0) { cleanup(tempdir2); }
     exit(1);
 }
 
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -50,16 +50,6 @@ string Strip0x(const string &hex)
     return hex;
 }
 
-bool endsWith(const string& str, const string& suffix)
-{
-    return str.size() >= suffix.size() && 0 == str.compare(str.size()-suffix.size(), suffix.size(), suffix);
-}
-
-bool startsWith(const string& str, const string& prefix)
-{
-    return str.size() >= prefix.size() && 0 == str.compare(0, prefix.size(), prefix);
-}
-
 void HelpAndQuit(cxxopts::Options options)
 {
     cout << options.help({""}) << endl;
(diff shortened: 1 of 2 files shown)


kd637xx:
### TODO
- [ ] handle exit and cleanup in the DiskPlotter class so it can be called from Python as well

src/cli.cpp
```diff
@@ -58,6 +70,30 @@ void HelpAndQuit(cxxopts::Options options)
     exit(0);
 }
 
+string filename = "plot.dat";
+string tempdir = ".";
+string tempdir2 = ".";
+
+void cleanup(const string& tmpdir)
+{
+    string glob = fs::path(tmpdir) / (filename + ".*.tmp");
```
arvidn, line 79: I don't think it's safe to do all of these things in a signal handler.
See https://man7.org/linux/man-pages/man7/signal-safety.7.html
notably, [readdir](https://www.man7.org/linux/man-pages/man3/readdir.3.html) is not in that list.
It may also be tempting for future patches to add more logic here, that wouldn't be allowed. I think a much safer and robust way of handling signals is for the signal handler to just communicate back to the main thread(s) that it should terminate, and run the regular clean-up code there.

src/cli.cpp
```diff
@@ -58,6 +70,30 @@ void HelpAndQuit(cxxopts::Options options)
...
+    cout << "Cleaning up " << glob << " .." << endl;        
+    for (const auto& f : fs::directory_iterator(tmpdir)) {
+        const auto fname = f.path().filename().string();
+        if (f.is_regular_file() && startsWith(fname, filename) && endsWith(fname, ".tmp")) {
```
arvidn, line 83: this assumption of filenames may be correct in 99% of the cases, but the time it *isn't* correct, it can have quite bad consequences. I think being accurate in which files to remove is important.
Consider running two plots in parallel for instance, sharing the same temp directory.

kd637xx: If two plots run in parallel in same tmpdir, `filename` will still be different, right?

arvidn: you might be right. It may have been a poor example. Maybe this rule is correct in more than 99% of cases, but you can still construct cases where it would delete files unrelated to the plot.

src/cli.cpp
kd637xx: Yes, cleanup should be done in DiskPlotter. I am thinking of ways to improve this so that is callable from Python (`chia plots create` process).

I am thinking the following methods would be good to have in DiskPlotter:

.stop()
.cleanup(phase = 1|2|..|all)

Additionally, would be great to have a way of pausing and resuming the plotting process or continuing partial plots, something like:

.pause()
.resume(plot_id, location) - resumes from last checkpoint (phase)

src/cli.cpp
kd637xx: I see what you mean: if using CLI, sure, if you name your first file `plot`, your second one `plot.x` and you kill the `plot` process, it will accidentally remove `plot.x.*` too. In that case, bad luck.

When using the DiskPlotter from `chia plots create`, which is the normal case, such scenario is not currently possible.

src/cli.cpp
kd637xx: Added `.stop()` and `.clean()` to DiskPlotter.

`.stop()` does not do anything currently but it should signal current phase to terminate (and I don't know how to do that atm)

PR #156, opened 2021-01-27
Upgrade to pybind 2.6.6 to avoid Python 3.9.0 bug

hoffmang9:
Also implement cibuildwheel 1.8.0 and migrate to a base of python 3.8


wjblanke (approved):
AOK

PR #157, opened 2021-01-27
Remove -mtune=native

hoffmang9:
Per @xorinox this is probably why current chiapos can be 10-20% slower than b17 based on local CPU.


wjblanke (approved):
awesome

arvidn:
the reason I added this was to ensure `__popcnt64()` would turn into the single-instruction on x86, rather than falling back to a software implementation. Perhaps new versions of GCC assume a recent enough x86 architecture to do this anyway.

hoffmang9:
We're going to package this into dev and have @xorinox test to see if the desired performance returns. If not, we'll revert.

arvidn:
to be clear, I don't doubt that this could cause a performance regression. I'm just suggesting it might be a good idea to verify that we still get the popcount instruction. There may be a better flag for instance, specifying a specific SSE version we require.

(although, I would be surprised if this flag causes a performance regression when built and run on the same machine).

PR #162, opened 2021-02-15
Ms.minor issues

mariano54:


src/prover_disk.hpp
```diff
@@ -46,11 +46,11 @@ struct plot_header {
...
 public:
-    // The costructor opens the file, and reads the contents of the file header. The table pointers
+    // The constructor opens the file, and reads the contents of the file header. The table pointers
     // will be used to find and seek to all seven tables, at the time of proving.
-    explicit DiskProver(std::string filename)
+    explicit DiskProver(const std::string& filename)
     {
-        struct plot_header header;
+        struct plot_header header{};
         this->filename = filename;
```
arvidn, line 54: if you instead would move the filename into the member variable here, callers of this function would be able to hand over ownership of the string, resulting in not copying the string at all. I think doing that is more ideomatic C++ these days, but it's not like it makes a difference here anyway.

src/prover_disk.hpp
```diff
@@ -254,18 +253,18 @@ class DiskProver {
 
         // This is the checkpoint at the beginning of the park
         uint16_t line_point_size = EntrySizes::CalculateLinePointSize(k);
-        uint8_t* line_point_bin = new uint8_t[line_point_size + 7];
+        auto* line_point_bin = new uint8_t[line_point_size + 7];
```
arvidn, line 256: these should probably be `std::unique_ptr<uint8_t[]>` too, (and created with `std::make_unique<uint8_t[]>(line_point_size + 7)`).

src/verifier.hpp
```diff
@@ -111,8 +111,8 @@ class Verifier {
                 new_ys.push_back(std::get<0>(results));
                 new_metadata.push_back(std::get<1>(results));
             }
-            for (size_t i = 0; i < new_ys.size(); i++) {
-                if (new_ys[i].GetSize() <= 0) {
+            for (auto & new_y : new_ys) {
```
arvidn, line 114: I think `auto const&` would be nicer here, to clarify that this loop doesn't mutate the elements

arvidn (approved):

arvidn:
mostly nit-picks

hoffmang9:
I'm going to merge this for a new tag but I'm going to assume @mariano54 can come back to the nits later if neccessary.

PR #172, opened 2021-02-23
Tmemowjb

wjblanke:
thread safe tables for FSE

### DIFF
--- a/src/encoding.hpp
+++ b/src/encoding.hpp
@@ -222,6 +222,7 @@ public:
             size_t err = FSE_buildDTable(dt, nCount.data(), maxSymbolValue, tableLog);
             if (FSE_isError(err)) {
                 throw FSE_getErrorName(err);
+            }
             tmc.DTAssign(R, dt);
         }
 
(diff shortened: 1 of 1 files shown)


hoffmang9 (approved):
+1

src/encoding.hpp
```diff
@@ -29,8 +29,68 @@
...
-std::map<double, FSE_CTable *> CT_MEMO = {};
-std::map<double, FSE_DTable *> DT_MEMO = {};
+#include <mutex>
+
+class TMemoCache {
+public:
+    ~TMemoCache() 
+    {
+        // Clean up global entries on destruction
+        std::lock_guard<std::mutex> l(memoMutex);
```
arvidn, line 39: this looks a bit suspicious. By the time this object is being destructed, no other thread has better hold a reference to it. Is this actually protecting against something that can happen?

Or was this mutex meant to be `static` as well, so multiple `TMemoCache` objects were meant to share it?

wjblanke: I don't think it is needed, I just had it to mirror the other functions. Its gone now.

hoffmang9 (approved):
heh on varname length

PR #173, opened 2021-03-04
Changes to support Apple M1

hoffmang9:


hoffmang9:
Also catches catch.hpp up

wjblanke (approved):
AOK

PR #174, opened 2021-03-04
Update calccrypto/uint128_t to 00714e6

hoffmang9:
https://github.com/calccrypto/uint128_t/commit/00714e6059d7d0d111b68a8fd9ee609094da3c0a


mariano54 (approved):
I can't say for sure whether these changes are safe or not.. But if it's passing plots on all platforms, I guess it's OK.

PR #176, opened 2021-03-08
Prevent parallel writes to rotational media

zmeyc:
During parallel plotting when copying multiple plots to dest simultaneously, if dest is HDD it'll result in slowdown and FS fragmentation.

This patch tries to detect whether dest is rotational media and if so, uses flock() to prevent multiple simultaneous writes to dest.

Linux only, on other platforms the behavior is unchanged.

Adds a log message notifying if parallel writing is enabled:
```
Starting plotting progress into temporary dirs: . and .
Final dir: . (parallel writing: disabled)
ID: 23023404333717545b0a6f0c0dde9710e4d3fe2d5cc6cc0a090a0b818bab0f17
Plot size is: 18
Buffer size is: 11MiB
Using 16 buckets
Using 2 threads of stripe size 4000
```


src/plotter_disk.hpp
```diff
@@ -398,11 +420,7 @@ class DiskPlotter {
             }
 
             if (!bRenamed) {
-#ifdef _WIN32
-                Sleep(5 * 60000);
-#else
-                sleep(5 * 60);
-#endif
+                Util::sleep_seconds(5 * 60);
```
arvidn, line 423: this could just be `std::this_thread::sleep_for(5min);`

for `min` you'd need `using namespace std::chrono_literals;` as well

src/disk_util.hpp
```diff
@@ -0,0 +1,189 @@
...
+        return true;
+#else
+        struct stat s;
+        memset(&s, 0, sizeof(s));
+
+        if (0 != stat(dir.c_str(), &s)) {
+            std::ostringstream err;
+            err << "Unable to find device name for dir " << dir << ": "
+                << strerror(errno);
+            throw std::runtime_error(err.str());
```
arvidn, line 54: it seems a bit gratuitous to throw here. The only thing you do with this exception is to turn it into a `false`. you might as well just return `false` directly (unless I'm misreading something)

arvidn: or `true` I suppose here, whatever is the current behavior

src/disk_util.hpp
```diff
@@ -0,0 +1,189 @@
...
+        try {
+            return !DiskUtil::IsParallelWritingPreferred(dir);
+        } catch (const std::runtime_error &err) {
+            std::cerr << err.what() << std::endl;
+            return false;
+        }
+    }
+}
+
+class DirectoryLock
```
arvidn, line 99: you would probably want this class to be non-copyable and non-movable (or movable, but you'd have to make it transfer ownership of the lock).

```
DirectoryLock(DirectoryLock const&) = delete;
```

src/disk_util.hpp
```diff
@@ -0,0 +1,189 @@
...
+        }
+        if (!UnlockDirectory(fd_, dirname_)) {
+            return false;
+        }
+        fd_ = -1;
+        return true;
+    }
+
+private:
+    static int LockDirectory(
```
arvidn, line 136: these two static functions should probably be free functions. They don't seem to require access to private fields of `DirectoryLock`, so they probably shouldn't have.

src/disk_util.hpp
zmeyc: I used exception because it's error path. Directory absence is unexpected. Also, errors from these exceptions are being logged and can be useful, i.e. "File not found" or "Access denied" etc.

src/disk_util.hpp
zmeyc: Fixed.

src/disk_util.hpp
zmeyc: Moved them to DiskUtil.

src/plotter_disk.hpp
zmeyc: Thanks! Fixed.

src/disk_util.hpp
zmeyc: @arvidn If you'd like to avoid exceptions in this function please let me know, I fixed / pushed the rest.

zmeyc: I realized that error handling is inconsistent with Lock/UnlockDir, I'll refactor this func to avoid exceptions.

zmeyc: Done. Pushed the fixes.

arvidn:
since there's no way to gracefully abort plotting (yet), it's likely to be killed with `SIGKILL`. Is there any risk the lock on the directory will be orphaned and just keep it locked in that case? or is the lock associated with the process and released automatically if the process is killed?

zmeyc:
@arvidn Yes, it's released automatically if process is killed.

zmeyc:
@arvidn Added logging of time taken to acquire the lock / repushed.

src/disk_util.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+    // For SSDs parallel writing is preferred.
+    // For HDDs it's slow and causes fragmentation.
+    static inline bool IsParallelWritingPreferred(const std::string &dir)
+    {
+#if defined(__APPLE__) || defined(_WIN32)
+        return true;
+#else
+        struct stat s;
+        memset(&s, 0, sizeof(s));
```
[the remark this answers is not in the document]
zmeyc: Fixed

src/disk_util.hpp
```diff
@@ -0,0 +1,189 @@
...
+#include <sys/sysmacros.h>
+#endif
+
+#include "util.hpp"
+
+namespace DiskUtil {
+
+    // For SSDs parallel writing is preferred.
+    // For HDDs it's slow and causes fragmentation.
+    static inline bool IsParallelWritingPreferred(const std::string &dir)
```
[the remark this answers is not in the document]
zmeyc: I was thinking about adding more logic in future like if it's a raid device (may prefer parallel writing even if rotational?)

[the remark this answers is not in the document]
zmeyc: Renamed to IsRotational for now.

src/plotter_disk.hpp
```diff
@@ -367,8 +375,29 @@ class DiskPlotter {
...
+
+                    std::chrono::steady_clock::time_point lock_start;
+                    if (should_lock) {
+                        std::cout << "Acquiring destdir lock" << std::endl;
+                        lock_start = std::chrono::steady_clock::now();
+                    }
+                    DirectoryLock dir_lock(final_dirname, should_lock);
+                    if (should_lock) {
+                        std::chrono::steady_clock::time_point lock_end = std::chrono::steady_clock::now();
+                        std::cout << "Lock acquired (took "
```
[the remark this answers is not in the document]
zmeyc: Done.

zmeyc:
@arvidn Added all the changes and cleaned up the commits.

src/disk_util.hpp
```diff
@@ -0,0 +1,197 @@
...
+#endif
+
+#include "util.hpp"
+
+namespace DiskUtil {
+
+    inline bool IsRotational(const std::string &dir)
+    {
+#if defined(__APPLE__) || defined(_WIN32)
+        return true;
```
arvidn, line 45: this should be `false` right? otherwise this would always lock on MacOS

src/disk_util.hpp
```diff
@@ -0,0 +1,197 @@
...
+            return true;
+        }
+
+        std::string line;
+        getline(file, line);
+
+        file.close();
+
+        int is_rotational = 1;
+        if (!line.empty() && line.front() == '0') {
```
arvidn, line 80: if this fails, the assumption should probably be conservative, that this is not rotational

src/disk_util.hpp
zmeyc: Yes, good catch. Inversed the logic in other places as well.

src/disk_util.hpp
zmeyc: Yes, fixed.

zmeyc:
Fixed merge conflicts.

zmeyc:
Please do not merge, found a problem with symlink (i.e. /mnt/dest -> /mnt/real-hdd) being incorrectly being detected as non-rotational media.

zmeyc:
Now works correctly with symlinks.

zmeyc:
@arvidn It was incorrect to zero-out minor device id. I changed the logic to generate a device path using both major and minor id, resolve the resulting symlink then traverse the final path up to the root looking for "queue/rotational". If not found, media is assumed to be non-rotational.

Example:

1) `/mnt/testlink` has major id 65, minor id 178

2) Generate path (symlink): "/sys/dev/block/65:178"

3) Resolve symlink to:
`/sys/devices/pci0000:00/0000:00:03.1/0000:09:00.0/host2/target2:1:4/2:1:4:0/block/sde/sde2`

4) Append "/queue/rotational" and check if file exists:

`/sys/devices/pci0000:00/0000:00:03.1/0000:09:00.0/host2/target2:1:4/2:1:4:0/block/sde/sde2/queue/rotational` - not found

5) Try root path until found:

`/sys/devices/pci0000:00/0000:00:03.1/0000:09:00.0/host2/target2:1:4/2:1:4:0/block/sde/queue/rotational` - found

6) If file contains '1' - media is rotational

zmeyc:
@arvidn Would it be possible to merge this PR? If any changes needed, please let me know.

src/disk_util.hpp
```diff
@@ -0,0 +1,229 @@
...
+        return fs::path(pathbuf.get());
+    }
+#endif
+
+    inline bool IsRotational(const std::string &dir)
+    {
+#if defined(__APPLE__) || defined(_WIN32)
+        return false;
+#else
+        struct stat s{};
```
arvidn, line 72: I think this function is linux specific, right?
It should be conditioned by `__linux__`, since right now it's for everything except MacOS and Windows.
I also think it should have a comment explaining what it does and why it works. Ideally with links to documentation stating that the assumptions made by this function are safe to make.

src/disk_util.hpp
```diff
@@ -0,0 +1,229 @@
...
+#include <sys/sysmacros.h>
+#endif
+
+#include "chia_filesystem.hpp"
+#include "util.hpp"
+
+namespace DiskUtil {
+
+#if !defined(__APPLE__) && !defined(_WIN32)
+    inline fs::path DevicePath(dev_t dev_id)
```
arvidn, line 45: this function also looks linux specific, it should be conditioned by `__linux__`.
Is the assumptions made by this function a part of the kernel or user space distro? Presumably the former, some links to documentations would be nice for readers.

src/disk_util.hpp
```diff
@@ -0,0 +1,229 @@
...
+        return !line.empty() && line.front() == '1';
+#endif
+    }
+
+    inline bool ShouldLock(const std::string &dir) {
+        return DiskUtil::IsRotational(dir);
+    }
+
+    inline int LockDirectory(
+        std::string dirname)
```
arvidn, line 125: is there a good reason for this to be `string` instead of `fs::path`?

src/plotter_disk.hpp
```diff
@@ -373,8 +381,14 @@ class DiskPlotter {
                 }
             } else {
                 if (!bCopied) {
+                    bool should_lock = DiskUtil::ShouldLock(final_dirname);
```
arvidn, line 384: this should probably be `const`

src/plotter_disk.hpp
```diff
@@ -153,9 +156,14 @@ class DiskPlotter {
         }
 #endif /* defined(_WIN32) || defined(__x86_64__) */
 
+        const char is_parallel_writing_enabled = !DiskUtil::ShouldLock(final_dirname);
```
arvidn, line 159: the name suggests that this should be a `bool`, yet, it's a `char`. Is there a good reason for this?

arvidn (commented):
Ideally there would be unit tests for as much of this as possible. A test that makes sure the directory lock works shouldn't be too difficult, and a test that ensures that the device ID returns something looking like what is expected also shouldn't be too hard.

Is there anything else that could be unit tested on CI?

PR #178, opened 2021-03-10
Replace Bits with uint128_t in phase 2

rostislav:
Improve CPU performance of phase 2 by 10% when using bitfield.

### DIFF
--- a/src/phase2.hpp
+++ b/src/phase2.hpp
@@ -53,6 +53,11 @@ Phase2Results RunPhase2(
     // An extra bit is used, since we may have more than 2^k entries in a table. (After pruning,
     // each table will have 0.8*2^k or fewer entries).
     uint8_t const pos_size = k;
+    uint8_t const pos_offset_size = pos_size + kOffsetSize;
+    uint8_t const write_counter_shift = 128 - (k + 1);
+    uint8_t const pos_offset_shift = write_counter_shift - pos_offset_size;
+    uint8_t const f7_shift = 128 - k;
+    uint8_t const t7_pos_offset_shift = f7_shift - pos_offset_size;
 
     std::vector<uint64_t> new_table_sizes(8, 0);
     new_table_sizes[7] = table_sizes[7];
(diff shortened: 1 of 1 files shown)


hoffmang9 (approved):
+1 - will merge once the k32 confirms same speed or better.

src/phase2.hpp
```diff
@@ -163,61 +167,49 @@ Phase2Results RunPhase2(
...
             std::tie(entry_pos, entry_offset) = index.lookup(entry_pos, entry_offset);
+            entry_pos_offset = (entry_pos << kOffsetSize) | entry_offset;
 
+            uint8_t bytes[16];
             if (table_index == 7) {
                 // table 7 is already sorted by pos, so we just rewrite the
                 // pos and offset in-place
+                uint128_t new_entry = (uint128_t)entry_f7 << f7_shift;
+                new_entry |= (uint128_t)entry_pos_offset << t7_pos_offset_shift;
+                Util::IntTo16Bytes(bytes, new_entry);
```
arvidn, line 200: is there a good reason to not do this byteswap in-place?
Also, if we always write a full 16 byte entries here, is there a good reason not to leave it in the host-endian?

rostislav: We don't always write 16 bytes here, the entry size is determined by `EntrySizes::GetMaxEntrySize`, which returns `3*k+1` bits rounded up to whole bytes. There's also a convention in the codebase that all integers are encoded as big-endian before being written to disk, including in `SortManager` which assumes this as well for sorting entries.

hoffmang9:
Early results are that this is 26% faster in phase 2 in bitfield. Merging and will then pull into chia-blockchain once we have final numbers.

Nicely done @rostislav

mariano54:
Nice. Does this put bitfield above non-bitfield, with 4GB of ram?

I would love to see a run of 4GB ram bitfield to compare. So then we can get finally rid of the old code.

hoffmang9:
@justinengland will be able to do a non bitfield run on same hardware during the day tomorrow to get a bottom line comparison between new bitfield and classic no bitfield.

PR #179, opened 2021-03-11
Retry FileDisk Open for should-be-existing files

L3Sota:
**Problem statement**

I found the "wait 5 minutes then try again" behaviour useful when plotting, since on multiple occasions I misjudged my remaining disk space and ran out of space. Currently, that behaviour doesn't exist when opening a file handle, which irrecoverably fails the plotting process if you move a temp file (say, to another disk to get some space back, so that other parallel plotting instances can continue).

**Solution**

Use the same "wait 5 minutes then retry" behaviour for `FileDisk` instances when they call `Open` in `Read` or `Write`, since they already `Open`ed successfully before (in the constructor).

**Notes**

I've held back from using this behaviour in the constructor, since that's the first time we encounter a particular file (and we aren't wasting a plotting run by failing at the beginning). It's also a unit test case, so I figured that I shouldn't change that behaviour.

**Testing**

1. Compiled and `mkdir tmp tmp2` as preparation
2. Ran `./ProofOfSpace create -s 16000 -u 0 -t ./tmp -d ./tmp -k 20`
3. `mv tmp/plot.dat.table2.tmp tmp2` during execution
4. After confirming the error message, moved it back `mv tmp2/plot.dat.table2.tmp tmp`

```
operation: create
Generating plot for k=20 filename=plot.dat id=022fb42c08c12de3a6af053880199806532e79515f94e83461612101f9412f9e

[. . .]

sorting table 3
sort time =  0.140 seconds. CPU (100.040%) Thu Mar 11 18:29:46 2021
Backpropagating on table 2
Could not open "./tmp/plot.dat.table2.tmp": No such file or directory. Retrying in five minutes.
scanned table 2
scanned time =  300.020 seconds. CPU (0.000%) Thu Mar 11 18:34:46 2021
sorting table 2
sort time =  0.151 seconds. CPU (99.290%) Thu Mar 11 18:34:46 2021
table 1 new size: 835133
Time for phase 2 = 301.046 seconds. CPU (0.310%) Thu Mar 11 18:34:46 2021
Wrote: 145

Starting phase 3/4: Compression from tmp files into "./plot.dat.2.tmp" ... Thu Mar 11 18:34:46 2021
Compressing tables 1 and 2

[. . .]

Time for phase 4 = 0.107 seconds. CPU (99.450%) Thu Mar 11 18:34:48 2021
Approximate working space used (without final file): 0.043 GiB
Final File size: 0.015 GiB
Total time = 306.523 seconds. CPU (2.470%) Thu Mar 11 18:34:48 2021
Copied final file from "./plot.dat.2.tmp" to "./tmp/plot.dat.2.tmp"
Removed temp2 file "./plot.dat.2.tmp"? 1
Renamed final file from "./tmp/plot.dat.2.tmp" to "./tmp/plot.dat"
```


src/disk.hpp
```diff
@@ -100,20 +100,35 @@ struct FileDisk {
         Open(false);
     }
 
-    void Open(bool read_only = true)
+    void Open(bool read_only = true, bool retry = false)
```
arvidn, line 103: This isn't directly an issue with this patch, but I think it's problematic to take multiple plain `bool`s like this. At the call site it will be very easy to misunderstand what each `true` or `false` mean.

I would suggest we instead pass a flag here, had have two named flags `read_only` and `retry`. That would improve the current situation with regards to readability at the call site as well.

src/disk.hpp
```diff
@@ -100,20 +100,35 @@ struct FileDisk {
...
-        f_ = ::_wfopen(filename_.c_str(), read_only ? L"r+b" : L"w+b");
+                f_ = ::_wfopen(filename_.c_str(), read_only ? L"r+b" : L"w+b");
 #else
-        f_ = ::fopen(filename_.c_str(), read_only ? "r+b" : "w+b");
+                f_ = ::fopen(filename_.c_str(), read_only ? "r+b" : "w+b");
 #endif
-        if (f_ == nullptr) {
-            throw InvalidValueException(
-                "Could not open " + filename_.string() + ": " + ::strerror(errno));
+                if (f_ == nullptr) {
```
arvidn, line 116: it seems to me that this code could be simplified by *only* checking the `retry` setting right here (i.e. not branch on it earlier). If it's not set, throw here, otherwise move along

arvidn (commented):
I haven't thought too much about the consequences of this. I can't think of any problems and it looks pretty good. I think there are a few improvements that could be made

src/disk.hpp
L3Sota: Sure, I can do that. I'm bad at C++ though, so I'd appreciate some pointers on where you want such flags to be defined. Maybe a nested struct in `FileDisk`?

```suggestion
    // Define right under `struct FileDisk {` as nested struct? Top-level in this file?
    struct FileOpenFlags {
        bool read_only = true;
        bool retry = false;
    }
    
    void Open(FileOpenFlags options)
```

src/disk.hpp
L3Sota: Yeah, it's not very compact. Thanks, I'll try your suggestion.

src/disk.hpp
L3Sota: Or maybe you meant something more like passing `uint8_t flags` into `Open`?

src/disk.hpp
```diff
@@ -97,39 +97,32 @@ struct FileDisk {
     explicit FileDisk(const fs::path &filename)
     {
         filename_ = filename;
-        Open(false);
+        Open();
     }
 
-    void Open(bool read_only = true, bool retry = false)
+    void Open(uint8_t flags = 0)
```
arvidn, line 103:
```suggestion
    void Open(uint8_t flags = readOnlyFlag)
```

arvidn (commented):
looks good apart from the default value for `flags`. Perhaps we don't rely on the default argument anywhere, but I think it's still a good idea to preserve the semantics

L3Sota:
Thank you for the second review! I've committed your suggestion.

src/disk.hpp
```diff
@@ -97,24 +97,32 @@ struct FileDisk {
     explicit FileDisk(const fs::path &filename)
     {
         filename_ = filename;
-        Open(false);
+        Open();
```
arvidn, line 100: actually. here's a case where we used to say read/write. but you changed it to rely on the default, so this changed to read-only now. since there is not "read_write" flag (it's just the absence of `ReadOnly`) maybe it would make sense to either:

1. invert the flag and have a `WriteFlag` instead of a `ReadOnlyFlag`. Then the default flag could be 0.
2. Change the default flag to be 0 and review all places we call `Open()` from to ensure we pass the right flag.
3. Pass `0` explicitly when we want read-write mode. This has the problem again that it's not clear at the call site what `0` means.

I would lean a bit towards (1), partly because I tend to think of the ability to write is the interesting property, we always open files in read-mode.

L3Sota: Agreed, I'll make that change. Just a moment.

L3Sota:
OK, checked that tests pass locally and did the manual file move thing again. Should work now.

src/disk.hpp
```diff
@@ -170,7 +178,7 @@ struct FileDisk {
 
     void Write(uint64_t begin, const uint8_t *memcache, uint64_t length)
     {
-        Open();
+        Open(writeFlag | retryOpenFlag);
```
arvidn, line 181: hm. this used to rely on the default argument, which was `true`, meaning it would open the file in read-only mode.
This looks like a bug. Perhaps it was never encountered because the file would always already be open at this point.
Anyway, as far as I can tell, your patch fixes it.

arvidn (approved):

arvidn:
@L3Sota thanks for your contribution!

src/disk.hpp
L3Sota: Oh, interesting. I didn't even realize that was a bug, but yeah, seems "fixed" now.

hellman_example/sort_on_disk.hpp
```diff
@@ -145,7 +145,7 @@ class FileDisk : public Disk {
         f_.rdbuf()->pubsetbuf(buf_, BUF_SIZE);
 
         if (!f_.is_open()) {
-            std::cout << "Fialed to open" << std::endl;
```
arvidn, line 148: this change appears to be conflicting with something happening in `main`. The files under `hellman_example` are also not actually used. afaik, they're left as a historical curiosity.

L3Sota: The line below it changed from a `string` to a `runtime_error`, no direct conflict. Resolved and pushed.

PR #182, opened 2021-03-16
CPU optimizations in phases 1 and 3

rostislav:


src/phase3.hpp
```diff
@@ -59,8 +59,8 @@ void WriteParkToFile(
     uint64_t writer = table_start + park_index * park_size_bytes;
     uint8_t *index = park_buffer;
 
-    Bits first_line_point_bits(first_line_point, 2 * k);
-    first_line_point_bits.ToBytes(index);
+    first_line_point <<= 128 - 2 * k;
+    Util::IntTo16Bytes(index, first_line_point);
```
arvidn, line 63: does the `park_buffer` also have the additional 7 trailing bytes?

src/phase3.hpp
```diff
@@ -399,10 +400,12 @@ Phase3Results RunPhase3(
...
 
             // Write the new position (index) and the sort key
-            Bits to_write = Bits(sort_key, right_sort_key_size);
-            to_write += Bits(index, k + 1);
+            uint128_t to_write = (uint128_t)sort_key << sort_key_shift;
+            to_write |= (uint128_t)index << index_shift;
 
-            L_sort_manager->AddToCache(to_write);
+            uint8_t bytes[16];
+            Util::IntTo16Bytes(bytes, to_write);
```
arvidn, line 407: this is safe without having the extra 7 bytes because it's all already aligned, right?

arvidn (approved):

src/util.hpp
```diff
@@ -339,6 +346,40 @@ namespace Util {
         double b = ldexp(frac, exp);
         return b;
     }
+
+#if defined(_WIN32) || defined(__x86_64__)
+    void CpuID(uint32_t leaf, uint32_t regs[4])
```
arvidn, line 351: passing an array as a parameter is misleading. this is the same as `uint32_t*`. i.e. the 4 is ignored.
If you want to enforce passing a C-array of 4, you can take a reference to the array, or use `std::array<uint32_t, 4>`

arvidn (approved):
just one nit

src/util.hpp
rostislav: Thanks, I'll change it to `uint32_t *`.

PR #183, opened 2021-03-20
use swap{16,32,64} on OpenBSD

timkuijsten:
This fixes a compile issue on OpenBSD. Tested on OpenBSD 6.8 stable.

### DIFF
--- a/src/util.hpp
+++ b/src/util.hpp
@@ -61,22 +61,10 @@ std::ostream &operator<<(std::ostream &strm, uint128_t const &v)
 #define bswap_16(x) _byteswap_ushort(x)
 #define bswap_32(x) _byteswap_ulong(x)
 #define bswap_64(x) _byteswap_uint64(x)
-#elif defined(__APPLE__)
-
-#include <libkern/OSByteOrder.h>
-
-#define bswap_16(x) OSSwapInt16(x)
-#define bswap_32(x) OSSwapInt32(x)
-#define bswap_64(x) OSSwapInt64(x)
-#elif defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
-# if defined(__OpenBSD__)
-#  include <machine/endian.h>
-# else
-#  include <sys/endian.h>
-# endif
-#define bswap_16(x) bswap16(x)
-#define bswap_32(x) bswap32(x)
-#define bswap_64(x) bswap64(x)
+#elif defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
+#define bswap_16(x) __builtin_bswap16(x)
+#define bswap_32(x) __builtin_bswap32(x)
+#define bswap_64(x) __builtin_bswap64(x)
 #else
 #include <byteswap.h>
 #endif


arvidn:
was this always broken or is this something that depends on version?

Generally, I would prefer to move towards using compiler intrinsics instead of OS facilities. We support fewer compilers than platforms. i.e. [__builtin_bswap16() et.al.](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) for GCC, Clang and Intel and [_byteswap_uint64() et.al.](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort?view=msvc-160) for Visual Studio. I believe that would simplify the code and cover more platforms.

timkuijsten:
> was this always broken or is this something that depends on version?

It looks like the bswap functions were never part of OpenBSD: https://github.com/openbsd/src/search?q=bswap16&type=commits.

> Generally, I would prefer to move towards using compiler intrinsics instead of OS facilities. We support fewer compilers than platforms. i.e. [__builtin_bswap16() et.al.](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) for GCC, Clang and Intel and [_byteswap_uint64() et.al.](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort?view=msvc-160) for Visual Studio. I believe that would simplify the code and cover more platforms.

Not sure if it is relevant here, but by default some builtin functions are explicitly disabled in favor of a libc implementation, mainly because of hardened memory management. Some default compiler flags are -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-valloc -fno-builtin-free -fno-builtin-strdup -fno-builtin-strndup.

```sh
$ cc -v test.c 2>&1 | fold -sw 95
OpenBSD clang version 10.0.1 
Target: amd64-unknown-openbsd6.8
Thread model: posix
InstalledDir: /usr/bin
 "/usr/bin/cc" -cc1 -triple amd64-unknown-openbsd6.8 -emit-obj -mrelax-all -disable-free 
-disable-llvm-verifier -discard-value-names -main-file-name test.c -mrelocation-model pic 
-pic-level 1 -pic-is-pie -mthread-model posix -mframe-pointer=all -relaxed-aliasing 
-fno-rounding-math -masm-verbose -mconstructor-aliases -munwind-tables -target-cpu x86-64 
-target-feature +retpoline-indirect-calls -target-feature +retpoline-indirect-branches 
-dwarf-column-info -fno-split-dwarf-inlining -debugger-tuning=gdb -v -resource-dir 
/usr/lib/clang/10.0.1 -fdebug-compilation-dir /home/tim -ferror-limit 19 -fmessage-length 0 
-fwrapv -D_RET_PROTECTOR -ret-protector -fgnuc-version=4.2.1 -fobjc-runtime=gnustep 
-fdiagnostics-show-option -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc 
-fno-builtin-valloc -fno-builtin-free -fno-builtin-strdup -fno-builtin-strndup -faddrsig -o 
/tmp/test-2caed2.o -x c test.c
clang -cc1 version 10.0.1 based upon LLVM 10.0.1 default target amd64-unknown-openbsd6.8
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/clang/10.0.1/include
 /usr/include
End of search list.
 "/usr/bin/ld" -e __start --eh-frame-hdr -Bdynamic -dynamic-linker /usr/libexec/ld.so -o a.out 
/usr/bin/../lib/crt0.o /usr/bin/../lib/crtbegin.o -L/usr/bin/../lib -L/usr/lib 
/tmp/test-2caed2.o -lcompiler_rt -lc -lcompiler_rt /usr/bin/../lib/crtend.o
```

rostislav:
I would also suggest using compiler builtins to reduce the amount of platform-specific code if they actually work on all relevant platforms. It's unlikely that `__builtin_bswap64` would get disabled by default for "hardening" purposes.

timkuijsten:
Ok, I have updated the code to use `__builtin_bswap{16,32,64}` on Apple and the BSDs. I've tested it on OpenBSD 6.8 and macOS 11.2.3. So with OpenBSD clang 10.0.1 and Apple clang 12.0.0.

arvidn:
I think this is *almost* a step in the right direction. It bothers me a bit that the use of the built-in is conditioned by *platform* rather than compiler. It's a compiler specific function. Conditioning it on compiler requires fixing the windows portion as well though.

arvidn:
This is what I had in mind. I think this would cover everything: https://github.com/Chia-Network/chiapos/pull/187

timkuijsten:
Tested a3c390f9 via main on OpenBSD, all works now, tnx!

PR #185, opened 2021-03-22
CPU and memory use optimizations

rostislav:
Increase size of bitfield index, use all available memory in phases 3 and 4.

### DIFF
--- a/src/phase2.hpp
+++ b/src/phase2.hpp
@@ -147,7 +147,7 @@ Phase2Results RunPhase2(
         // * sort by pos
 
         auto sort_manager = std::make_unique<SortManager>(
-            memory_size / 2,
+            table_index == 2 ? memory_size : memory_size / 2,
             num_buckets,
             log_num_buckets,
             uint16_t(entry_size),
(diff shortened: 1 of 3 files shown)


src/phase2.hpp
```diff
@@ -147,7 +147,7 @@ Phase2Results RunPhase2(
         // * sort by pos
 
         auto sort_manager = std::make_unique<SortManager>(
-            memory_size / 2,
+            table_index == 2 ? memory_size : memory_size / 2,
```
arvidn, line 150: I think this warrants a comment

arvidn (commented):
looks good to me. Did you run this through a heap profiler to ensure it doesn't exceed the limit?

src/phase2.hpp
rostislav: Good idea, I will provide one

rostislav:
> looks good to me. Did you run this through a heap profiler to ensure it doesn't exceed the limit?

I did not use a heap profiler, but I traced all calls to constructors and destructors of `SortManager` to ensure we stay within the limit.

hoffmang9 (approved):
Decent speed up in parallel plotting and single path plotting

PR #186, opened 2021-03-23
Add simple progress percentage display

aqk:


src/progress.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef SRC_CPP_PROGRESS_HPP_
+#define SRC_CPP_PROGRESS_HPP_
+
+#include <iostream>
+
+void progress(int phase, int n, int max_n) {
+    float p = (100.0 / 4) * ((phase - 1.0) + (1.0 * n / max_n));
+    std::cout << "Progress: " << p << std::endl;
```
arvidn, line 8: I'm a fan of:

```
std::cout << "\rProgress: " << p << "  ";
std::cout.flush();
```

In case this is printed frequently, that won't make your terminal scroll so much

arvidn: you may also want `<< std::setprecision(2) << (p * 100.0) << "% "`

aqk: That's a good idea. I'll make that change.

aqk: I think we still need a line break after p, so that you don't see this:

```
Progress: 42.0F1 complete, time: 0.011 seconds. CPU (151.67%) Mon Mar 22 23:18:05 2021
```

on the same line

mariano54:
I think we should do opt-in (or opt-out) with a flag, otherwise other tools like plotman will not work properly, because they rely on old logs.

rostislav:
I agree we should introduce a flag to show progress to avoid polluting output for CLI users.

hoffmang9 (approved):
With the flag added, this looks good.

PR #187, opened 2021-03-23
use compiler intrinsics for byte swapping

arvidn:
instead of platform specific functions/macros.
This is simpler and more robust since we support fewer compilers than platforms

### DIFF
--- a/src/util.hpp
+++ b/src/util.hpp
@@ -54,31 +54,24 @@ std::ostream &operator<<(std::ostream &strm, uint128_t const &v)
 
 #endif
 
-/* Platform-specific byte swap macros. */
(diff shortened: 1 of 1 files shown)


timkuijsten:
Tested and build successfully with OpenBSD 6.8, clang 10.0.1 and Apple macOS 11.2.3, clang 12.0.0.

rostislav (commented):
Just a couple of nits: we can also remove "else include byteswap.h" since this header is GNU-specific and this case is already covered. Another one is about commit message: these are compiler builtins and not "intrinsics", which are different things.

hoffmang9 (approved):
Still need to test on other BSDs

PR #189, opened 2021-03-25
don't fall off the end of the deltas into garbage

wjblanke:
./ProofOfSpace prove ffffff11239639109fd2ecdfff3f51457f9f805a5294fbee660a5f89c9e11683

### DIFF
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -326,8 +326,10 @@ private:
         std::vector<uint8_t> deltas =
             Encoding::ANSDecodeDeltas(bit_mask, encoded_size, kCheckpoint1Interval, kC3R);
         std::vector<uint64_t> p7_positions;
+        bool surpassed_f7 = false;
         for (uint8_t delta : deltas) {
             if (curr_f7 > f7) {
+                surpassed_f7 = true;
                 break;
             }
             curr_f7 += delta;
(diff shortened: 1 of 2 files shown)


mariano54 (approved):

mariano54 (changes requested):
Actually this is still not working for me.
 ./ProofOfSpace prove fffffa2b647d4651c500076d7df4c6f352936cf293bd79c591a7b08e43d6adfb
 plot id: plot-k22-cc01d66ab532c2293e44a3a5f14325b92991ed4d5bf5a4c69efb9f766e0d0326.plot

mariano54:
I'm working on a fix now.

hoffmang9 (approved):
+1

hoffmang9 (approved):
Seems like my approval didn't stick?

mariano54:
Not sure why the test fail... But I tried chia plots check and it seemed good

mariano54 (approved):

src/prover_disk.hpp
```diff
@@ -326,8 +326,10 @@ class DiskProver {
         std::vector<uint8_t> deltas =
             Encoding::ANSDecodeDeltas(bit_mask, encoded_size, kCheckpoint1Interval, kC3R);
         std::vector<uint64_t> p7_positions;
+        bool surpassed_f7 = false;
         for (uint8_t delta : deltas) {
             if (curr_f7 > f7) {
+                surpassed_f7 = true;
```
arvidn, line 332: perhaps I'm missing something, but it looks like we could just return `p2_positions` here, remove the `surpassed_f7` boolean and return an empty vector unconditionally at the end of the function.

src/prover_disk.hpp
```diff
@@ -337,12 +339,19 @@ class DiskProver {
...
+            // we have an actual solution for f7.
             if ((int64_t)curr_p7_pos >= (int64_t)((c1_index + 1) * kCheckpoint1Interval) - 1 ||
-                curr_f7 >= (((uint64_t)1) << k)) {
-                return p7_positions;
+                curr_f7 >= (((uint64_t)1) << k) - 1) {
+                break;
             }
         }
+        if (!surpassed_f7) {
+            return std::vector<uint64_t>();
```
arvidn, line 353: I'm pretty sure this could even just be `return {};`

src/prover_disk.hpp
mariano54: Yeah that is simpler..

PR #191, opened 2021-03-26
Feat: measure time to copy from -2 to -d

L3Sota:
Resolves: #168

### DIFF
--- a/src/plotter_disk.hpp
+++ b/src/plotter_disk.hpp
@@ -364,6 +364,7 @@ public:
 
         bool bCopied = false;
         bool bRenamed = false;
+        Timer copy;
         do {
             std::error_code ec;
             if (tmp_2_filename.parent_path() == final_filename.parent_path()) {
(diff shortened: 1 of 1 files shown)


mariano54 (approved):

src/plotter_disk.hpp
```diff
@@ -364,6 +364,7 @@ class DiskPlotter {
 
         bool bCopied = false;
         bool bRenamed = false;
+        Timer copy;
```
arvidn, line 367: did you deliberately include possible wait time in this timer?

arvidn (approved):

src/plotter_disk.hpp
L3Sota: Yes. The phase 1/2/3/4 and overall timers are all wall clock timers and include any delays due to "5 min. later" retries, process suspension, etc., so I figured this is a consistent approach to measuring the "copy phase" at the end.

PR #192, opened 2021-03-27
hellman_example|src: Throw proper exceptions

xdustinface:


mariano54 (approved):
Thanks! The hellman source is not used BTW, so it's not worth spending much time on it.

jwest75674:
@mariano54
I am diving into and studying chiapos and chia-blockchain with the hope to eventually contribute. In most cases it's been a smooth process of research, but specific to /hellman_example, I have not been able to find the history of decision making and am hoping you might be able to shine a light on the same.

Though experimental currently, the potential to reduce final plot sizes seems like it would be in-demand.

What was the thought process that lead to the hellman process/src not being used?

hoffmang9:
We left the Hellman attack as low hanging fruit for our proof of space contest but no one took advantage of it. There are two steps there. You would probably want to write a separate little tool to compress existing plots and then look at the time/effort trade offs of running that during the plotting process. I suspect that it's most efficient to complete plots and then slowly move through Hellman attacking them.

PR #193, opened 2021-03-27
lib: Simplify FSE_decompress_usingDTable

xdustinface:


mariano54 (approved):

xdustinface:
I just realized this is changing a third party library :see_no_evil:

hoffmang9:
Yeah - I looked into fixing our import here and found we'd forked it...

PR #200, opened 2021-03-30
make the read and write disk caches smaller

arvidn:
4 MiB -> 1 MiB. These are multiplied by the number of buckets.
the larger cache size is believed to not contribute to any performance improvements.


rostislav (approved):
I agree that these cache sizes should be reduced.

PR #201, opened 2021-03-31
Update pos_constants.hpp

xorinox:
Allow for up to 1024 buckets

### DIFF
--- a/src/pos_constants.hpp
+++ b/src/pos_constants.hpp
@@ -26,7 +26,7 @@ const uint32_t kOffsetSize = 10;
 // Max matches a single entry can have, used for hardcoded memory allocation
 const uint32_t kMaxMatchesSingleEntry = 30;
 const uint32_t kMinBuckets = 16;
-const uint32_t kMaxBuckets = 128;
+const uint32_t kMaxBuckets = 1024;
 
 // During backprop and compress, the write pointer is ahead of the read pointer
 // Note that the large the offset, the higher these values must be


arvidn:
looks good to me, but I'd like to defer approval to someone who knows where that limit comes from.

arvidn:
@rostislav do you know?

rostislav:
Similarly to PR #202, I would suggest making the commit message more descriptive.
Regarding the reason for the original limit, it might be related to the limit on open file handles in Windows, but I'm not sure.

arvidn:
indeed. it seems we would need [this](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setmaxstdio?view=msvc-160) for windows.

> C run-time I/O now supports up to 8,192 files open simultaneously at the low I/O level. This level includes files opened and accessed using the _open, _read, and _write family of I/O functions. By default, up to **512** files can be open simultaneously at the stream I/O level. This level includes files opened and accessed using the fopen, fgetc, and fputc family of functions. The limit of 512 open files at the stream I/O level can be increased to a maximum of **8,192** by use of the _setmaxstdio function.
> Because stream I/O-level functions, such as fopen, are built on top of the low I/O-level functions, the maximum of 8,192 is a hard upper limit for the number of simultaneously open files accessed through the C run-time library.
>

mariano54:
I think this is a good idea and hopefully should allow us to drastically reduce memory requirements.

1. The reason I set it to 128 was the following: The idea was that splitting writes into too many buckets would severely slow down HDD plotting because it would constantly be seeking through these writes. However this was just a theory, i never actually tested anything more than 128.
2. If we do this change, I think we should tweak the amount of memory allocated to the sortmanager for writes by increasing it, and decrease the amount of memory going to the reads. This means that more stuff will be cached before writing to disk. Hopefully that should improve HDD performance. That might increase total memory requirements slightly.
3. Do you have any data for performance xorinox? I think we should test this on an HDD.
4. I'm curious about what the minimum ram requirements will be changed to, with different -u sizes. We should confirm that the code actually runs with the decreased -b parameters, and doesn't crash

xorinox:
A lot of buckets certainly not the idea for spinner plotting, but for SSD I found increasing from 128 to 512 beneficial, approx 10% plotting throughput gains on multi process plotting, more importantly, can comfortably plot a no speed loos with -b 1024, 8 concurrent procs on a machine with only 32G of RAM.   Thanks to reducing the disk read/write buffers from 4 to 1MB per bucket file.  My testing was all done on 0.12.41   but I am on my way to check out the differences on chaipos commit a614cee546841ff08a5ae248f7b09a0062f6576d also

wjblanke (commented):
I think we need to check the system ulimit (platform specific?) and use that to determine max buckets. Otherwise there will be failures later when the file handles are exhausted and a lot of support questions. Some machines have a very low setting. If we add that it is fine by me.

PR #202, opened 2021-03-31
Update plotter_disk.hpp

xorinox:
Setting the rlimit also on Linux was necessary to make it work with 1024 buckets plus changing the number from 600 to 4096.

### DIFF
--- a/src/plotter_disk.hpp
+++ b/src/plotter_disk.hpp
@@ -73,12 +73,11 @@ public:
         bool show_progress = false)
     {
         // Increases the open file limit, we will open a lot of files.
-#ifndef _WIN32
-        struct rlimit the_limit = {600, 600};
+        struct rlimit the_limit = {4096, 4096};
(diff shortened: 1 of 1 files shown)


src/plotter_disk.hpp
```diff
@@ -73,12 +73,11 @@ class DiskPlotter {
         bool show_progress = false)
     {
         // Increases the open file limit, we will open a lot of files.
-#ifndef _WIN32
```
arvidn, line 76: you need to put these `#ifndef` back. Note the `n` in there, this condition is to omit this code on windows, because windows doesn't have rlimits. (this is why this patch fails to build on windows, [here](https://github.com/Chia-Network/chiapos/pull/202/checks?check_run_id=2233385499#step:8:235))

arvidn (changes requested):
looks good apart from the windows build issue.

rostislav:
I would also suggest to make the commit message more descriptive, e.g. "Increase file descriptor limit"

PR #205, opened 2021-04-08
Prevent chiapos prover crashing for bad plots

mariano54:
I replicated segfaults on bad plots (scrambling or removing 0.5% of the plot for example).
In the cases of corrupted data (0.5% of the plot is randomly scrambled), the prover should still be able to return correct responses > 99% of the time, so this means faulty plots are still farmable. If the size of the plot is messed with, however, it wont be farmable, it will just throw (but no longer segfault). 

* Check that the stream is good right after seeking and reading. This prevents further issues down the line, by throwing immediately.
* When we read deltas size that are too large, throw immediately. The most common symptom of a bad plot is that the delta size is too large, which caused issues in decoding.

### DIFF
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -166,17 +166,25 @@ int main(int argc, char *argv[]) try {
         HexToBytes(challenge, challenge_bytes);
 
         DiskProver prover(filename);
-        vector<LargeBits> qualities = prover.GetQualitiesForChallenge(challenge_bytes);
-        for (uint32_t i = 0; i < qualities.size(); i++) {
-            k = prover.GetSize();
-            uint8_t *proof_data = new uint8_t[8 * k];
-            LargeBits proof = prover.GetFullProof(challenge_bytes, i);
-            proof.ToBytes(proof_data);
-            cout << "Proof: 0x" << Util::HexStr(proof_data, k * 8) << endl;
-            delete[] proof_data;
-        }
-        if (qualities.empty()) {
-            cout << "No proofs found." << endl;
+        try {
+            vector<LargeBits> qualities = prover.GetQualitiesForChallenge(challenge_bytes);
+            for (uint32_t i = 0; i < qualities.size(); i++) {
+                k = prover.GetSize();
+                uint8_t *proof_data = new uint8_t[8 * k];
+                LargeBits proof = prover.GetFullProof(challenge_bytes, i);
+                proof.ToBytes(proof_data);
+                cout << "Proof: 0x" << Util::HexStr(proof_data, k * 8) << endl;
+                delete[] proof_data;
+            }
+            if (qualities.empty()) {
+                cout << "No proofs found." << endl;
+                exit(1);
+            }
+        } catch (const std::exception& ex) {
+            std::cout << "Error proving. " << ex.what() << std::endl;
+            exit(1);
+        } catch (...) {
+            std::cout << "Error proving. " << std::endl;
             exit(1);
         }
     } else if (operation == "verify") {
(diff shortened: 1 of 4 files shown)


src/prover_disk.hpp
```diff
@@ -241,6 +242,32 @@ class DiskProver {
...
+            throw std::runtime_error("goodbit!=0 after seeking to" + std::to_string(seek_location));
+        }
+    }
+
+    void safeRead(std::ifstream& disk_file, uint8_t* target, uint64_t size) {
+        if ((disk_file.rdstate() & std::ifstream::goodbit) != 0) {
+            throw std::invalid_argument("safeRead: rovided input stream goodbit != 0.");
+        }
+        disk_file.read(reinterpret_cast<char*>(target), size);
+        if ((disk_file.rdstate() & std::ifstream::goodbit) != 0) {
```
rostislav, line 265: This actually doesn't work. `std::ifstream::goodbit` is 0 on my machine, so we never enter the `if`. We should instead test `std::ifstream::failbit | std::ifstream::badbit` to catch errors.

src/prover_disk.hpp
```diff
@@ -241,6 +242,32 @@ class DiskProver {
...
+        disk_file.seekg(seek_location);
+
+        if ((disk_file.rdstate() & std::ifstream::goodbit) != 0) {
+            std::cout << "goodbit, failbit, badbit, eofbit: " <<   (disk_file.rdstate() & std::ifstream::goodbit) <<  (disk_file.rdstate() & std::ifstream::failbit) << (disk_file.rdstate() & std::ifstream::badbit) << (disk_file.rdstate() & std::ifstream::eofbit) << std::endl;
+            throw std::runtime_error("goodbit!=0 after seeking to" + std::to_string(seek_location));
+        }
+    }
+
+    void safeRead(std::ifstream& disk_file, uint8_t* target, uint64_t size) {
+        if ((disk_file.rdstate() & std::ifstream::goodbit) != 0) {
```
rostislav, line 261: Probably it's extraneous to check the state before `read`. Also see the next comment about `goodbit` being 0.

src/prover_disk.hpp
```diff
@@ -241,6 +242,32 @@ class DiskProver {
...
     std::vector<uint64_t> C2;
 
+    // Using this method instead of simply seeking will prevent segfaults that would arise when
+    // continuing the process of looking up qualities.
+    void safeSeek(std::ifstream& disk_file, uint64_t seek_location) {
+        if ((disk_file.rdstate() & std::ifstream::goodbit) != 0) {
+            throw std::invalid_argument("safeSeed: provided input stream goodbit != 0.");
+        }
+
+        disk_file.seekg(seek_location);
```
rostislav, line 252: Seeking beyond end-of-file is not an error. I think it's necessary to only wrap calls to `read`, but not to `seekg`.

src/prover_disk.hpp
```diff
@@ -241,6 +242,32 @@ class DiskProver {
...
+
+        if ((disk_file.rdstate() & std::ifstream::goodbit) != 0) {
+            std::cout << "goodbit, failbit, badbit, eofbit: " <<   (disk_file.rdstate() & std::ifstream::goodbit) <<  (disk_file.rdstate() & std::ifstream::failbit) << (disk_file.rdstate() & std::ifstream::badbit) << (disk_file.rdstate() & std::ifstream::eofbit) << std::endl;
+            throw std::runtime_error("goodbit!=0 after seeking to" + std::to_string(seek_location));
+        }
+    }
+
+    void safeRead(std::ifstream& disk_file, uint8_t* target, uint64_t size) {
```
rostislav, line 260: The name should rather be `SafeRead` to fit the style used for other functions' names.

rostislav (changes requested):
It's good that we're now failing gracefully if plot data is invalid. We should also test with a truncated plot file to ensure we handle unexpected EOF correctly.

src/cli.cpp
```diff
@@ -241,25 +249,30 @@ int main(int argc, char *argv[]) try {
...
+                        cout << "quality: " << quality << endl;
+                        cout << "Proof verification suceeded. k = " << static_cast<int>(k) << endl;
+                        success++;
+                    } else {
+                        cout << "Proof verification failed." << endl;
+                    }
+                    delete[] proof_data;
                 }
-                delete[] proof_data;
+            } catch (const std::exception error) {
```
rostislav, line 273: Probably a missing `&` here.

src/prover_disk.hpp
mariano54: Are you saying seeking beyond the end will not trigger a badbit or failbit? It does under my tests.
Basically an invalid pointer read from disk (bad plot) will cause seek to set badbit or failbit.

src/prover_disk.hpp
mariano54: OK, ill check for fail and badbits

src/prover_disk.hpp
rostislav: We should check the error bits set after `.read()` in `safeRead`. Checking them once again here is not necessary. I'm still reasonably sure that `.seekg()` will not set error bits by itself, since seeking beyond EOF is a valid operation. Which OS did you test it on?

mariano54:
There are situations in which seekg can fail:
https://www.cplusplus.com/reference/istream/istream/seekg/

src/prover_disk.hpp
```diff
@@ -241,6 +242,26 @@ class DiskProver {
     std::vector<uint64_t> table_begin_pointers;
     std::vector<uint64_t> C2;
 
+    // Using this method instead of simply seeking will prevent segfaults that would arise when
+    // continuing the process of looking up qualities.
+    static void SafeSeek(std::ifstream& disk_file, uint64_t seek_location) {
+        disk_file.seekg(seek_location);
+
+        if (disk_file.bad() || disk_file.fail()) {
```
rostislav, line 250: This is equivalent to checking only `.fail()`, as it returns true when either of `failbit` or `badbit` is set.

src/prover_disk.hpp
```diff
@@ -241,6 +242,26 @@ class DiskProver {
...
+            std::cout << "goodbit, failbit, badbit, eofbit: " <<   (disk_file.rdstate() & std::ifstream::goodbit) <<  (disk_file.rdstate() & std::ifstream::failbit) << (disk_file.rdstate() & std::ifstream::badbit) << (disk_file.rdstate() & std::ifstream::eofbit) << std::endl;
```
rostislav, line 251: I'd suggest to split the too long line.

src/prover_disk.hpp
```diff
@@ -241,6 +242,26 @@ class DiskProver {
...
+            throw std::runtime_error("badbit or failbit after seeking to" + std::to_string(seek_location));
```
rostislav, line 252: Space is needed after "to".

src/prover_disk.hpp
```diff
@@ -241,6 +242,26 @@ class DiskProver {
...
+        }
+    }
+
+    static void SafeRead(std::ifstream& disk_file, uint8_t* target, uint64_t size) {
+        disk_file.read(reinterpret_cast<char*>(target), size);
+
+        if (disk_file.bad() || disk_file.fail()) {
+            std::cout << "goodbit, failbit, badbit, eofbit: " <<   (disk_file.rdstate() & std::ifstream::goodbit) <<  (disk_file.rdstate() & std::ifstream::failbit) << (disk_file.rdstate() & std::ifstream::badbit) << (disk_file.rdstate() & std::ifstream::eofbit) << std::endl;
+            throw std::runtime_error("badbit or failbit after reading size" + std::to_string(size));
```
rostislav, line 261: Similar comments apply as in `SafeSeek`.

rostislav (changes requested):

src/prover_disk.hpp
mariano54: done

src/prover_disk.hpp
mariano54: done

src/prover_disk.hpp
mariano54: done

rostislav (approved):
I've added file position to the error message in case of read error - this should be useful for debugging if it happens.
Otherwise, it's good for me.

PR #206, opened 2021-04-08
Improve logging

xdustinface:
Allow to disable logging and also to use other `std::ostream` instances as destination for the logs like in the CLI example here 8e5005d.


wjblanke:
return out << tfm::format("%s seconds. CPU (%s%%)", wall_clock_ms, cpu_ratio);

Is it possible to remove the tinyformat dependency and switch this to using sprintf or something more basic?

xdustinface:
> Windows doesn't have this header.

@wjblanke No OS has it but it gets added in 58e0e4fdd0ac55152eca5e29a1ca65bc14dd71c3 which also works fine for windows, see https://github.com/Chia-Network/chiapos/runs/2298802999.

It seems like thats only an issue for the wheels build but i am not familiar with that python wheel stuff so if anyone has an idea what might lead to it not finding it there, tell me!

wjblanke:
Anything windows related on python is dicey at best. There just isn't a lot of support from the python community for Windows.

Since tinyformat is only used in one place, can we just replace it with a more basic implementation and remove the dependency? It would save a lot of time for the build team.

> > Windows doesn't have this header.
> 
> @wjblanke No OS has it but it gets added in [58e0e4f](https://github.com/Chia-Network/chiapos/commit/58e0e4fdd0ac55152eca5e29a1ca65bc14dd71c3) which also works fine for windows, see https://github.com/Chia-Network/chiapos/runs/2298802999.
> 
> It seems like thats only an issue for the wheels build but i am not familiar with that python wheel stuff so if anyone has an idea what might lead to it not finding it there, tell me!

hoffmang9:
It needs to be added here - https://github.com/Chia-Network/chiapos/blob/main/setup.py#L109

xdustinface:
> It needs to be added here - https://github.com/Chia-Network/chiapos/blob/main/setup.py#L109

Cool, that worked! But i removed it from cmake and just copied over the file into `lib/include`.

hoffmang9:
Having yet another unmaintainable direct copy/fork isn't something I'm ok with.

Since it is used so sparingly, can you complete this PR without adding a new dependency?

xdustinface:
There is no standard way (i know about) of formatting like this with custom types. So i guess i could add some `ToString` methods and calls to them here and there and somehow use `sprintf` or so or i could try to hack out some dependency-free solution which works with custom types but it honestly feels saver to just use a mature third party library which exists for years. Its sad that the the windows wheels build doesn't work with the cmake integration that easy but im pretty sure this can be fixed i just have other stuff to do and don't have the time to dive into this that deep.. i have no windows/python-wheel experience. Since this dependency is a single header file, is it really that bad? Keep in mind that it can be replaced with `std::format` (with some changes to the format strings oviously) as soon as you use C++20, see https://en.cppreference.com/w/cpp/utility/format/format.

So, i would rather not want to spend more time on this PR but im totally fine adding a commit to make it dependency free or to fix the windows-wheel if someone wants to provide it.

xdustinface:
- Rebased on `main` to cleanup the new `std::cout` logs from #205 
- Added https://github.com/Chia-Network/chiapos/compare/a2a16e0aeb662130903ff5c182512327070059cf..355e1cc9d59926a64df1add4a6c3dbc3e5bd41d5#diff-563591a10c31a403aa309b1130739fdb2784d0d683cae13069c89fc81c6f7c30L407-R414 as fix because before this PR `std::endl` triggered the flush. 
- Added: 355e1cc9d59926a64df1add4a6c3dbc3e5bd41d5 (I can move it to a separate PR if you wish)

PR #207, opened 2021-04-19
code cleanup: remove useless code logic

BinacsLee:
Obviously, when `entry_y == prev_y`, the value of `entry_y - prev_y` will be zero.

And we can use:

```c++
            deltas_to_write.push_back(entry_y - prev_y);
```

instead of:

```c++
            if (entry_y == prev_y) {
                deltas_to_write.push_back(0);
            } else {
                deltas_to_write.push_back(entry_y - prev_y);
            }
```

Thanks for your review!


arvidn (approved):

hoffmang9:
Thank you for the PR!

PR #208, opened 2021-04-19
Optimize code in phase1.hpp

683280:
memcpy(&(right_writer_buf[i * entry_size_bytes]), to_write, 16);
This memcpy function will be called 4194304000 times
we can use:
	Util::IntTo16Bytes(&right_writer_buf[i * entry_size_bytes], entry);
instead of:
	Util::IntTo16Bytes(to_write, entry);
	memcpy(&(right_writer_buf[i * entry_size_bytes]), to_write, 16);
Delete memcpy

### DIFF
--- a/src/phase1.hpp
+++ b/src/phase1.hpp
@@ -558,13 +558,11 @@ void* F1thread(int const index, uint8_t const k, const uint8_t* id, std::mutex* 
         // to increase CPU efficency.
         f1.CalculateBuckets(x, loopcount, f1_entries.get());
         for (uint32_t i = 0; i < loopcount; i++) {
-            uint8_t to_write[16];
             uint128_t entry;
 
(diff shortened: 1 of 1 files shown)


arvidn:
I think this is a reasonable change, mostly for simplifying the code.
Did you measure a speed improvement? (I suspect the compiler would apply a very similar transformation as this patch already)

arvidn (approved):

rostislav (approved):

tcinbis:
With the PR branch rebased onto the current main branch in this repo, the timing is `102s` on main vs `110s` on 683280:main. Note these numbers come from running the benchmark command from the readme once. So take it with a grain of salt :sweat_smile:

wjblanke (approved):
aok

PR #211, opened 2021-04-21
Move final file if in same file system, not only in same directory

yEinFallsLos:
Use `std::filesystem::equivalent` instead of just comparing the parent directories to determine if tmp_2_file can be moved.
Prevents useless IO activity and speeds up plotting if temp_2_file and final_file are in the same file system.


aqk:
Hello - thanks for your fix. We do want to move to C++ stdlib calls, but have had to move slowly because of the number of platforms we support.

@wjblanke will this work on the oldest macos we currently support?

.gitignore
```diff
@@ -25,3 +25,4 @@ build
 .mypy_cache
 *.whl
 venv
+/.vs
```
arvidn, line 28: was this deliberate?

src/plotter_disk.hpp
```diff
@@ -369,7 +369,9 @@ class DiskPlotter {
         Timer copy;
         do {
             std::error_code ec;
-            if (tmp_2_filename.parent_path() == final_filename.parent_path()) {
+            bool same_filesystem =
+                fs::equivalent(tmp_2_filename.parent_path(), final_filename.parent_path(), ec);
```
arvidn, line 373: the name `same_filesystem` make this sound like a broader check than it is. The parent paths may refer to different directories, but still live within the same filesystem.

In order to prefer a rename, when that's possible, I think the most reliable way is to try renaming, and if it fails, fall back to copying.

.gitignore
yEinFallsLos: Visual Studio added it, and I thought since .vscode is already in there I can include it

src/plotter_disk.hpp
yEinFallsLos: `fs::equivalent` checks whether the paths resolve to the same file system entity, so I think `same_filesystem` is an appropriate name for that variable

Of course we could just try and error, but I think this a more elegant solution.

yEinFallsLos: I am sorry, my bad. I misread the documentation of `fs::equivalent` as that it checks whether the two paths resign in the same filesystem.
I updated my pull request to try a rename before falling back to a copy.

wjblanke:
> Hello - thanks for your fix. We do want to move to C++ stdlib calls, but have had to move slowly because of the number of platforms we support.
> 
> @wjblanke will this work on the oldest macos we currently support?

I think the question is whether Gulrak supports it. If it is compiling on MacOS should be OK

hoffmang9:
Also note there is an update to Gulrak - https://github.com/Chia-Network/chiapos/pull/259

PR #222, opened 2021-04-29
Reopen files upon read or write failure

mreid-moz:
If an attempt to read or write a temp file fails, close the
file before sleeping, then re-open it after sleeping.

This improves robustness when plotting with removable drives.
Previously, if a drive is unmounted/remounted during the plotting
process, the open files never recover and you get stuck in a loop
of:

`Only read 0 of <size> bytes at offset <offset> from "<filename>" with length <length>. Error 1. Retrying in five minutes.`

This change closes and re-opens files and can recover from (some?)
transient read and write problems. Specifically, if you have a
removable drive, and you unplug it during phase 2, it will recover
and continue when you plug the drive back in.

### DIFF
--- a/src/disk.hpp
+++ b/src/disk.hpp
@@ -175,7 +175,11 @@ struct FileDisk {
                 std::cout << "Only read " << amtread << " of " << length << " bytes at offset "
                           << begin << " from " << filename_ << "with length " << writeMax
                           << ". Error " << ferror(f_) << ". Retrying in five minutes." << std::endl;
+                // Close, Reopen, and re-seek the file
+                Close();
+                bReading = false;
                 std::this_thread::sleep_for(5min);
+                Open(retryOpenFlag);
             }
         } while (amtread != length);
     }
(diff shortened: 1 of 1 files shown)


src/disk.hpp
```diff
@@ -175,7 +175,11 @@ struct FileDisk {
                 std::cout << "Only read " << amtread << " of " << length << " bytes at offset "
                           << begin << " from " << filename_ << "with length " << writeMax
                           << ". Error " << ferror(f_) << ". Retrying in five minutes." << std::endl;
+                // Close, Reopen, and re-seek the file
```
arvidn, line 178: I think someone reading this code a year from now may wonder why we close and re-open the file again. I think this is a good place for a comment explaining the problem this solved.

src/disk.hpp
```diff
@@ -208,7 +212,11 @@ struct FileDisk {
                 std::cout << "Only wrote " << amtwritten << " of " << length << " bytes at offset "
                           << begin << " to " << filename_ << "with length " << writeMax
                           << ". Error " << ferror(f_) << ". Retrying in five minutes." << std::endl;
+                // Close, Reopen, and re-seek the file
```
arvidn, line 215: this is also a good place explaining the problem it solved

mreid-moz:
Updated the comment to reflect the problem being solved.

arvidn (approved):

PR #223, opened 2021-05-01
Add the verbose option to the sort managers

jespino:


python-bindings/chiapos.cpp
```diff
@@ -68,12 +69,26 @@ PYBIND11_MODULE(chiapos, m)
...
+            py::arg("filename"),
+            py::arg("k"),
+            py::arg("memo"),
+            py::arg("id"),
+            py::arg("buffmegabytes"),
+            py::arg("num_buckets"),
+            py::arg("stripe_size"),
+            py::arg("num_threads"),
+            py::arg("nobitfield"),
+            py::arg("verbose") = true);
```
arvidn, line 91: is there a good reason to enable verbose by default?

src/phase1.hpp
```diff
@@ -598,7 +598,8 @@ std::vector<uint64_t> RunPhase1(
     uint32_t const stripe_size,
     uint8_t const num_threads,
     bool const enable_bitfield,
-    bool const show_progress)
+    bool const show_progress,
+    bool verbose)
```
arvidn, line 602: I think 3 "naked" bools as function arguments exceed the limit of good readability.

I think it would make sense to replace these `bools` with a single `flags` argument instead. That way the call sites will tell you which features are enabled.

arvidn (commented):
variables that do not need to be mutable should be declared `const` (including parameters, such as `verbose`).

python-bindings/chiapos.cpp
jespino: Yes, the reason is "backward compatibility", the current behavior is to show the lines with "Block X uniform sort" text. this `verbose = false` is for avoid printing that. If we don't care about preserve the old behavior, I'm happy to set it to false by default.

src/phase1.hpp
jespino: I'm ok with that, I can make another PR for that, or I can try to do the changes here (But I prefer to only change 1 thing at a time per PR).

jespino:
@arvidn any feedback on my comments? I'm wating for you feedback to make the changes needed.

src/phase1.hpp
arvidn: In that case I would prefer that PR land first

jespino: Sure, no problem, I'll do that or first.

jespino:
@arvidn I added again the changes on top of the main branch, now using the new flag, let me know if you see something that needs to change there :)

jespino:
@arvidn I going to close this PR, feel free to reopen it again and ask me to do the fixes that are needed, but I'm cleaning my PRs list.

Issue #224, opened 2021-05-01
Print Target/Final Directory early in the logs


grayfallstown:
Final Directory not printed in running plotter logfiles, only printed at the end.

I was asked to show the Final Directory in [ChiaPlotStatus](https://grayfallstown.github.io/Chia-Plot-Status/). The problem is the final directory is logged at the end of phase 4, so I cannot show the final directory of running plotting processes.

Pull Request incoming.

PR #225, opened 2021-05-01
Print Target/Final Directory early in the logs refs #224

grayfallstown:


grayfallstown:
Any chance this can be merged?

hoffmang9:
This will get merged but it will take a few more days as there are other things that need to get done first and the merge here does not yet get it into chia-blockchain

wjblanke (approved):
aok

PR #226, opened 2021-05-02
Update disk.hpp

zhaoyangwx:
Avoid stuck in windows disc cache

### DIFF
--- a/src/disk.hpp
+++ b/src/disk.hpp
@@ -201,6 +201,7 @@ struct FileDisk {
             }
             amtwritten =
                 ::fwrite(reinterpret_cast<const char *>(memcache), sizeof(uint8_t), length, f_);
+            _commit(_fileno(f_));
             writePos = begin + amtwritten;
             if (writePos > writeMax)
                 writeMax = writePos;


hoffmang9:
GitHub has a new process that requires a team member to trigger Actions. They are now running.

src/disk.hpp
```diff
@@ -201,6 +206,9 @@ struct FileDisk {
             }
             amtwritten =
                 ::fwrite(reinterpret_cast<const char *>(memcache), sizeof(uint8_t), length, f_);
+#ifdef _WIN32
+            _commit(_fileno(f_));
```
arvidn, line 210: it's not obvious that this is an improvement. This will block until the bytes have made it to the disk, right?
Could you provide some explanation of the symptom you're experiencing that this solves?
Ideally it would be explained in a comment here as well, to let future readers of the code understand why this call is here

zhaoyangwx: The symptom is seen in https://github.com/Chia-Network/chia-blockchain/issues/2337

arvidn: could you provide measurements demonstrating the improvement with this patch?

arvidn: the reason I feel cautious about this is that you really want disk *writes* to happen in parallel with your program executing, which means sticking the data in kernel buffers that are flushed to disk at some later time, while the program keeps running. As far as I can tell, this will effectively stop the program to wait for the data to make it all the way down to the platter (presumably surviving a power outage).

Generally, you save a lot of time to keep running instead.

src/disk.hpp
```diff
@@ -219,6 +216,14 @@ struct FileDisk {
                 std::this_thread::sleep_for(5min);
             }
         } while (amtwritten != length);
+    #ifdef _WIN32
+        // Flush Windows Cache after few cycles
+        flush_cyclecount += 1;
+        if (flush_cyclecount >= flush_cyclelimit){
```
arvidn, line 222: This makes more sense. However, if this turns out to give some people worse performance, they really should have a configuration option to disable it or change this limit.

The limit should probably be specified in number-of-bytes written to the file too.

src/disk.hpp
arvidn: the [documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/commit?view=msvc-160) doesn't actually say whether the call blocks until the data has been flushed or not.

PR #236, opened 2021-05-13
Migrating to flags instead of booleans parameters for show_progress and nobitfield

jespino:


jespino:
@arvidn here is the PR converting that arguments to flags. I'm not 100% sure if the new `phases.hpp` file is the right approach, le me know if you want me to move that to other place.

src/phases.hpp
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SRC_CPP_PHASES_HPP_
+#define SRC_CPP_PHASES_HPP_
+
+#pragma once
+enum phase_flags {
```
arvidn, line 19: you could specify the underlying type to match here, `enum phase_flags : uint8_t {`

arvidn (approved):

src/phases.hpp
jespino: Done :)

wjblanke (approved):
aok

PR #237, opened 2021-05-14
Add Process ID to log files

grayfallstown:
Motivation:

A user notices a plotting process fails to continue or that he configured it wrong while reading the logs or using a monitoring solution for plotting processes.
The user needs to find the plotting process that belongs to the log file.
Currently users try to match their configurations like the used temp drives, threads and bucket sizes from the log file to the command lines of running processes in Task Manager / System Viewer / HTOP so that they can kill the misbehaving process.
Having the Process ID as part of the log file greatly simplifies this process.


arvidn (approved):

tinodj:
I was doing this with lsof :) but, I agree PID number in the log file is great.

wjblanke (approved):
aok

PR #245, opened 2021-05-17
Common template for inline vector implementation.

lemur73:
Allows simplier using inline vectors of different sizes and types.

Signed-off-by: Ostap Slavking <lemur73@bk.ru>
If you like my commits cheer me with some chia here:
xch16es77qggpwgzucqmmvhvgcueckmt9g3e7exa46vhmucz7z8j2a2spm7c7e


arvidn:
this mostly looks like an improvement. However, when you make the element type a template argument I think you're going a bit too far. your `InlineVector` implementation does not support arbitrary element types. Fortunately, we don't need arbitrary types. I would suggest just leaving it hard coded as `uint64_t`.

One thing these classes need more urgently are tests though.

src/bits.hpp
```diff
@@ -27,61 +27,63 @@
 // 64 * 2^16. 2^17 values, each value can store 64 bits.
 #define kMaxSizeBits 8388608
 
-// A stack vector of length 5, having the functions of std::vector needed for Bits.
-struct SmallVector {
-    typedef uint16_t size_type;
+template <class item_type, class size_type_arg, unsigned capacity>
+class InlineVector {
+  public:
+    typedef size_type_arg size_type;
```
arvidn, line 33: the way this spells nowadays is:
```suggestion
using size_type = size_type_arg;
```

src/bits.hpp
```diff
@@ -27,61 +27,63 @@
...
 
-    void push_back(uint64_t value) { v_[count_++] = value; }
+    void push_back(item_type value) {
+        assert(count_ < capacity);
+        v_[count_++] = value;
+    }
 
-    SmallVector& operator=(const SmallVector& other)
-    {
+    InlineVector& operator=(const InlineVector& other) {
```
arvidn, line 52: while you're at it, this function should probably be ref-qualified.
```suggestion
    InlineVector& operator=(const InlineVector& other) & {
```

src/bits.hpp
```diff
@@ -27,61 +27,63 @@
...
-    size_type count_;
-};
-
-// A stack vector of length 1024, having the functions of std::vector needed for Bits.
-// The max number of Bits that can be stored is 1024 * 64
-struct ParkVector {
-    typedef uint32_t size_type;
-
-    ParkVector() noexcept { count_ = 0; }
+    InlineVector& operator=(const std::vector<item_type>& other) {
```
arvidn, line 58: and so should this probably:
```suggestion
    InlineVector& operator=(const std::vector<item_type>& other) & {
```

src/bits.hpp
```diff
@@ -27,61 +27,63 @@
...
+    item_type v_[capacity];
     size_type count_;
 };
 
+// A stack vector of length 10, having the functions of std::vector needed for Bits.
+typedef InlineVector<uint64_t, uint8_t, 10> SmallVector;
+
+// A stack vector of length 2048, having the functions of std::vector needed for Bits.
+// The max number of Bits that can be stored is 2048 * 64
+typedef InlineVector<uint64_t, uint16_t, 2048> ParkVector;
```
arvidn, line 85: the modern way (and more readable in my opinion) is to use a `using` declaration.
```suggestion
using ParkVector = InlineVector<uint64_t, uint16_t, 2048>;
```

lemur73:
> this mostly looks like an improvement. However, when you make the element type a template argument I think you're going a bit too far. your `InlineVector` implementation does not support arbitrary element types. Fortunately, we don't need arbitrary types. I would suggest just leaving it hard coded as `uint64_t`.

Well, I have some use of this template argument in my next commits. Just prefer to keep uploads smaller to make it easy to understand and review.

> One thing these classes need more urgently are tests though.

I'll upload tests.

src/bits.hpp
```diff
@@ -27,61 +27,63 @@
...
+    void resize(const size_type n) {
+        assert(n <= capacity);
+        count_ = n;
+    }
+
+    size_type max_size() const { return capacity; }
+
+  private:
+    item_type v_[capacity];
     size_type count_;
```
arvidn, line 77: If you make this change, you don't have to define the constructor

```suggestion
    size_type count_ = 0;
```

src/bits.hpp
lemur73: > the modern way (and more readable in my opinion) is to use a `using` declaration.

Unfortunately, this doesn't work. Gives an error:
```
~/chiapos_up/src/bits.hpp:79: error: expected nested-name-specifier
 using InlineVector<uint64_t, uint8_t, 10> SmallVector;
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/bits.hpp
lemur73: > One thing these classes need more urgently are tests though.

Updated with fixes from your comments and test.
PTAL

src/bits.hpp
arvidn: try this instead:
```
using ParkVector = InlineVector<uint64_t, uint16_t, 2048>;
```

lemur73: That worked, thanks.
Updated and uploaded.

src/bits.hpp
```diff
@@ -27,61 +27,61 @@
...
 #define kMaxSizeBits 8388608
 
-// A stack vector of length 5, having the functions of std::vector needed for Bits.
-struct SmallVector {
-    typedef uint16_t size_type;
-
-    SmallVector() noexcept { count_ = 0; }
-
-    uint64_t& operator[](const uint16_t index) { return v_[index]; }
+template <class item_type, class size_type_arg, unsigned capacity>
```
arvidn, line 30: Here's my argument against making `item_type` a template parameter:

As it stands in your patch, it's not correct. The template parameter cannot be an arbitrary type, since they're all constructed up-front they must be default constructable. In order to make this more correct, you would need something like:
```
static_assert(std::is_integral_t<item_type>, "InlineVector inly supports integral types");
```

Which brings me to my next point. We don't currently need this additional generality/complexity. I don't think it makes sense to make things more general/complex just for the sake of it (in fact, I would argue that keeping everything as specific as possible makes for easier-to-read programs).

Now, you say you'll need this template parameter in your next PR. So, let's punt this change until that PR.

arvidn:
I think everything looks good apart from my last comment

src/bits.hpp
lemur73: > The template parameter cannot be an arbitrary type, since they're all constructed up-front they must be default constructable. 

Yeah, I figured that out already in my next patch. Let's see if you like it ;)
Added static assert for now.

Issue #234, opened 2021-05-11
Chia plotting stops/hangs up at some point with CPU thread at 100%


tinodj:
Hi, 

this could be hw related, but however, maybe software can/should get out of the loop and print some error message, so one can easier investigate what the problem is. And also, stop waisting resources, rather terminate if can't make any progress. 

At our side it stops at random places, in the 3rd phase mostly. I have all the data ( p3.t2.sort_bucket_xxx.tmp ) saved at the point it stopped, if you want me to enable some logging/debugging from here on. Last message I had was about sorting bucket 53, in table 1/2 in phase 3. 

I took a look into the code very briefly, and I can see several while-s going on here, so some of them might not be terminating properly. Please note - no disk activity is happening, while the processor thread is running at 100%. 

If you suspect hardware, what it would be? RAM / DISK or something else? 

If you give me some instructions how to run more verbose/debugging version, I can do that and maybe that will nail down this problem for me and for others. 

This seems related:
https://github.com/Chia-Network/chia-blockchain/discussions/3228#discussioncomment-687990

tinodj:
Great @mhurliman that you will do some debugging. Please share if you find something. I will have no time before weekend to do any debugging on this, unfortunately. We have one plotter running without any troubles but we got these troubles on the second one (similar config to the first one). For now we suspect faulty RAM.

arvidn:
when plotting behaves strange, it's a good idea to run [memtest](https://www.memtest86.com/).

mhurliman:
Thanks for the response arvidn - I'll get it a run!

room101-dev:
**> when plotting behaves strange, it's a good idea to run [memtest](https://www.memtest86.com/).**

That's the common retort here, but what I find is even better, is try another NVME I run 3 on my plotter, seeing the same problems, and read this site and all the comments were RAM, and I did run memtest++ at boot, just for the hell, but in deeper analysis I found it was only one NVME having the problem, and looking into 'sudo nvme smart-log /dev/xxx' I found some troubling stat's, seems that around 25% of warrany TBW this stuff starts happening, I know on this particular NVME that sky-rocketing temp +90C was the first thing, followed by hangs, and then what didn't hang blew out at post with 'mis-match' on plot completion, which means you lose that plot.

[...]

PR #248, opened 2021-05-20
Always on error reset position/pointer

tinodj:
If an error occurs, the resulting value of the file-position indicator for the stream is unspecified. https://pubs.opengroup.org/onlinepubs/007904975/functions/fwrite.html

And in the code above if error occurs with 0 bytes written (full disk) it will not reset the pointer (writePos will still be equal to begin), however it need to be reseted. 

Otherwise this causes https://github.com/Chia-Network/chiapos/issues/234 - in phase3, when this bucket is read, it goes into endless loop.

### DIFF
--- a/src/disk.hpp
+++ b/src/disk.hpp
@@ -193,8 +193,9 @@ struct FileDisk {
 #endif
         // Seek and write from memcache
         uint64_t amtwritten;
+        bool errorOnWrite;
         do {
-            if ((bReading) || (begin != writePos)) {
+            if ((bReading) || (begin != writePos) || errorOnWrite) {
 #ifdef _WIN32
                 _fseeki64(f_, begin, SEEK_SET);
 #else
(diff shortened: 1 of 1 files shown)


src/disk.hpp
```diff
@@ -193,8 +193,9 @@ struct FileDisk {
 #endif
         // Seek and write from memcache
         uint64_t amtwritten;
+        bool errorOnWrite;
```
arvidn, line 196: instead of introducing a new variable, you could also set `bReading = true` on line 219 (222 in your patch).

I think regardless this warrants a comment to remind readers that the write position is unspecified when an error occurs

arvidn (approved):
great find!

tinodj:
Thanks. I was/am not sure what bReading means in context of this write method, and for easier code reading I introduced new variable. I can see bReading is used in Read function, and it is clear to me how it is used there. Maybe here, setting to false, should be actually setting to true on line 219/222 as you suggested. Please feel free to fix the best way you think it is.

arvidn:
`bReading` indicates whether the last operation was a read or write. if it is different than the current operation, we need to update the file position, since our cached copy of the file position will be outdated.

I agree that it's not ideal to repurpose this to trigger the `seek()`. I still think a comment is important though.

arvidn:
perhaps a more intuitive patch to fix this would be to set `writePos` to `UINT64_MAX` on failure. That models the behavior of the underlying file position being undefined.

src/disk.hpp
wjblanke: I think errorOnWrite needs to be initialized here, since it is checked in the conditional below

wjblanke:
Fixed with PR 267 based on your PR tinodj. Thank u!

PR #250, opened 2021-05-21
Explicitly track used entries in UniformSort

lukasstockner:
This slightly increases memory usage, but makes the sort significantly faster.

For example, a quick test with k=27 goes from 452sec to 360sec.


src/uniformsort.hpp
```diff
@@ -44,18 +36,19 @@ namespace UniformSort {
...
-        uint64_t const memory_len = Util::RoundSize(num_entries) * entry_len;
+        uint64_t const rounded_entries = Util::RoundSize(num_entries);
         auto const swap_space = std::make_unique<uint8_t[]>(entry_len);
         auto const buffer = std::make_unique<uint8_t[]>(BUF_SIZE);
         uint64_t bucket_length = 0;
         // The number of buckets needed (the smallest power of 2 greater than 2 * num_entries).
         while ((1ULL << bucket_length) < 2 * num_entries) bucket_length++;
-        memset(memory, 0, memory_len);
+
+        auto const is_used = std::make_unique<uint8_t[]>(rounded_entries);
```
arvidn, line 46: it seems you only use 1 out of these 8 bits. You can make this use 1/8th the amount of memory.
You could use the bitfield class we already have: https://github.com/Chia-Network/chiapos/blob/main/src/bitfield.hpp

src/uniformsort.hpp
```diff
@@ -67,35 +60,35 @@ namespace UniformSort {
             buf_size--;
             // First unique bits in the entry give the expected position of it in the sorted array.
             // We take 'bucket_length' bits starting with the first unique one.
-            uint64_t pos =
-                Util::ExtractNum(buffer.get() + buf_ptr, entry_len, bits_begin, bucket_length) *
-                entry_len;
+            uint64_t idx =
+                Util::ExtractNum(buffer.get() + buf_ptr, entry_len, bits_begin, bucket_length);
+            uint64_t mem_ptr = idx * entry_len;
```
arvidn, line 65: I think it's misleading to call this `_ptr`. It's not a pointer, it's an offset

src/uniformsort.hpp
```diff
@@ -67,35 +60,35 @@ namespace UniformSort {
...
-             pos += entry_len) {
-            if (!IsPositionEmpty(memory + pos, entry_len)) {
+        for (uint64_t idx = 0; entries_written < num_entries && idx < rounded_entries; idx++) {
+            if (is_used[idx]) {
                 // We've found an entry.
                 // write the stored entry itself.
                 memcpy(
                     memory + entries_written * entry_len,
-                    memory + pos,
+                    memory + idx * entry_len,
```
arvidn, line 91: in the loop above, you introduce a separate variable (`mem_ptr`), adding `entry_len` each iteration through the loop. I assume you did this to avoid the multiplication inside the loop. Is there a reason not to do the same thing here?

src/uniformsort.hpp
lukasstockner: I based the naming on `buf_ptr`, which is also an offset instead of a pointer. I agree that it's misleading, I'll change it.

src/uniformsort.hpp
lukasstockner: I mostly introduced `mem_ptr` to simplify the code, I'd expect the compiler to automatically create the temporary variable instead of multiplying each time.
But yeah, making it explicit is probably a good idea.

src/uniformsort.hpp
lukasstockner: Ah, yes, that's much nicer. There's a tiny speed hit (back up to 365sec), but that's okay I think.

src/uniformsort.hpp
arvidn: I would also expect the compiler to replace the multiplication in the loop. Either way is fine, I just wanted to make sure it was deliberate.

ftchann:
This uses way more disk space for me.

arvidn:
@ftchann that seems unintuitive. Could you post before- and after measurements?

hoffmang9:
And please use k=32 for both runs.

ftchann:
I think i made a mistake, about that.

src/uniformsort.hpp
wjblanke: Well, maybe hold off on this as POPCNT is not supported by all CPUs and we use this even on non bitfield

arvidn: this patch doesn't use the bitfield to count bits though.

arvidn: but it's a good point I hadn't thought of. You can't use all parts of the bitfield class unconditionally

JelloHuang:
it seem that the patch does not reduce time on k=32 when i tested

ConnorChristie:
I just gave this a try with Swar plot manager and it keeps crashing about halfway through 3 parallel plots.

PR #265, opened 2021-05-26
Fix inverted bitfield flag in python bindings.

MrHacky:
This is a fix for a bug that was introduced in a6fce5d485edb4b3b216389b203571392dccbff6.


wjblanke (approved):
aok

PR #280, opened 2021-06-09
Update CMakeLists.txt

risner:
FreeBSD has a port for pybind11, so use it when on FreeBSD instead of downloading a new copy (CMakeLists.txt change).
Typo in setup.py (vdf instead of pos).
Makefile for hellman attack directory. Does this code compile? I failed to get it to compile with clang 11.


hoffmang9 (approved):
looks good and thanks.

PR #291, opened 2021-07-02
Add parallel reads to GetQualitiesForChallenge

cccat6:
Speed up quality check by parallel `GetQualitiesForChallenge` outer loop.
Especially help NFS drive. It is able to speed up 25% overall proofing time in my testing.


cccat6:
I didn't get it. Why it doesn't work on some of the platforms and the error is so strange.

cccat6:
I will check in my side

PR #296, opened 2021-07-19
Option to revert to sequential read

emlowe:
There are some systems (primarily macos+exfat) where the parallel read features results in very long lookup times. This makes the parallel feature the default, but adds the ability to disable it and revert back to sequential reads. Made need more tests.

### DIFF
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -83,7 +83,7 @@ int main(int argc, char *argv[]) try {
     string id = "022fb42c08c12de3a6af053880199806532e79515f94e83461612101f9412f9e";
     bool nobitfield = false;
     bool show_progress = false;
-    bool disable_parallel_read = false;
+    bool parallel_read = true;
     uint32_t buffmegabytes = 0;
 
     options.allow_unrecognised_options().add_options()(
@@ -103,8 +103,8 @@ int main(int argc, char *argv[]) try {
         cxxopts::value<uint32_t>(buffmegabytes))(
         "p, progress", "Display progress percentage during plotting",
         cxxopts::value<bool>(show_progress))(
-        "disable_parallel_read", "Do not use parallel reads - revert back to sequential",
-        cxxopts::value<bool>(disable_parallel_read)->default_value("false"))(
+        "parallel_read", "Set to false to use sequential reads",
+        cxxopts::value<bool>(parallel_read)->default_value("true"))(
         "help", "Print help");
 
     auto result = options.parse(argc, argv);
@@ -180,7 +180,7 @@ int main(int argc, char *argv[]) try {
             for (uint32_t i = 0; i < qualities.size(); i++) {
                 k = prover.GetSize();
                 uint8_t *proof_data = new uint8_t[8 * k];
-                LargeBits proof = prover.GetFullProof(challenge_bytes, i, disable_parallel_read);
+                LargeBits proof = prover.GetFullProof(challenge_bytes, i, parallel_read);
                 proof.ToBytes(proof_data);
                 cout << "Proof: 0x" << Util::HexStr(proof_data, k * 8) << endl;
                 delete[] proof_data;
@@ -262,7 +262,7 @@ int main(int argc, char *argv[]) try {
                 vector<LargeBits> qualities = prover.GetQualitiesForChallenge(hash.data());
 
                 for (uint32_t i = 0; i < qualities.size(); i++) {
-                    LargeBits proof = prover.GetFullProof(hash.data(), i, disable_parallel_read);
+                    LargeBits proof = prover.GetFullProof(hash.data(), i, parallel_read);
                     uint8_t *proof_data = new uint8_t[proof.GetSize() / 8];
                     proof.ToBytes(proof_data);
                     cout << "i: " << num << std::endl;
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -83,6 +83,7 @@ int main(int argc, char *argv[]) try {
     string id = "022fb42c08c12de3a6af053880199806532e79515f94e83461612101f9412f9e";
     bool nobitfield = false;
     bool show_progress = false;
+    bool disable_parallel_read = false;
     uint32_t buffmegabytes = 0;
 
     options.allow_unrecognised_options().add_options()(
@@ -102,6 +103,8 @@ int main(int argc, char *argv[]) try {
         cxxopts::value<uint32_t>(buffmegabytes))(
         "p, progress", "Display progress percentage during plotting",
         cxxopts::value<bool>(show_progress))(
+        "disable_parallel_read", "Do not use parallel reads - revert back to sequential",
+        cxxopts::value<bool>(disable_parallel_read)->default_value("false"))(
         "help", "Print help");
 
     auto result = options.parse(argc, argv);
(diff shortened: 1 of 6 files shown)


hoffmang9 (approved):
lgtm

src/prover_disk.hpp
```diff
@@ -635,11 +639,18 @@ class DiskProver {
...
     // recursively calling GetInputs for table 4.
-    std::vector<Bits> GetInputs(uint64_t position, uint8_t depth)
+    std::vector<Bits> GetInputs(uint64_t position, uint8_t depth, std::ifstream* disk_file=nullptr)
     {
-        // Create individual file handles to allow parallel processing
-        std::ifstream disk_file(filename, std::ios::in | std::ios::binary);
-        uint128_t line_point = ReadLinePoint(disk_file, depth, position);
+        uint128_t line_point;
+
+        if (!disk_file) {
```
xdustinface, line 646:
```suggestion
        if (disk_file == nullptr) {
```
Clearer to read imo.

src/prover_disk.hpp
```diff
@@ -649,14 +660,23 @@ class DiskProver {
             ret.emplace_back(xy.first, k);   // x
             return ret;
         } else {
-            auto left_fut=std::async(std::launch::async, &DiskProver::GetInputs,this, (uint64_t)xy.second, (uint8_t)(depth - 1));
-            auto right_fut=std::async(std::launch::async, &DiskProver::GetInputs,this, (uint64_t)xy.first, (uint8_t)(depth - 1));
-            std::vector<Bits> left = left_fut.get();  // y
-            std::vector<Bits> right = right_fut.get();  // x
+            std::vector<Bits> left, right;
+            if (!disk_file) {
```
xdustinface, line 664:
```suggestion
            if (disk_file == nullptr) {
```

src/prover_disk.hpp
```diff
@@ -649,14 +660,23 @@ class DiskProver {
...
+                // no disk_file, so we do parallel reads here
+                auto left_fut=std::async(std::launch::async, &DiskProver::GetInputs,this, (uint64_t)xy.second, (uint8_t)(depth - 1), nullptr);
+                auto right_fut=std::async(std::launch::async, &DiskProver::GetInputs,this, (uint64_t)xy.first, (uint8_t)(depth - 1), nullptr);
+                left = left_fut.get();  // y
+                right = right_fut.get();  // x
+            } else 
+            {
```
xdustinface, line 671:
```suggestion
            } else {
```

src/prover_disk.hpp
```diff
@@ -635,11 +639,18 @@ class DiskProver {
     // all of the leaves (x values). For example, for depth=5, it fetches the position-th
     // entry in table 5, reading the two back pointers from the line point, and then
     // recursively calling GetInputs for table 4.
-    std::vector<Bits> GetInputs(uint64_t position, uint8_t depth)
+    std::vector<Bits> GetInputs(uint64_t position, uint8_t depth, std::ifstream* disk_file=nullptr)
```
xdustinface, line 642:
```suggestion
    std::vector<Bits> GetInputs(uint64_t position, uint8_t depth, std::ifstream* disk_file = nullptr)
```

src/prover_disk.hpp
```diff
@@ -219,7 +219,11 @@ class DiskProver {
             }
 
             // Gets the 64 leaf x values, concatenated together into a k*64 bit string.
-            std::vector<Bits> xs = GetInputs(p7_entries[index], 6);
+            std::vector<Bits> xs;
+            if (disable_parallel_reads)
+                xs = GetInputs(p7_entries[index], 6, &disk_file); // Passing in a disk_file disabled the parallel reads
+            else
+                xs = GetInputs(p7_entries[index], 6);
```
xdustinface, line 226:
```suggestion
            if (disable_parallel_reads) {
                xs = GetInputs(p7_entries[index], 6, &disk_file); // Passing in a disk_file disabled the parallel reads
            } else {
                xs = GetInputs(p7_entries[index], 6);
            }
```

Feels just safer..

src/cli.cpp
```diff
@@ -102,6 +103,8 @@ int main(int argc, char *argv[]) try {
         cxxopts::value<uint32_t>(buffmegabytes))(
         "p, progress", "Display progress percentage during plotting",
         cxxopts::value<bool>(show_progress))(
+        "disable_parallel_read", "Do not use parallel reads - revert back to sequential",
```
xdustinface, line 106: No short version?

src/prover_disk.hpp
```diff
@@ -201,7 +201,7 @@ class DiskProver {
     // Given a challenge, and an index, returns a proof of space. This assumes GetQualities was
     // called, and there are actually proofs present. The index represents which proof to fetch,
     // if there are multiple.
-    LargeBits GetFullProof(const uint8_t* challenge, uint32_t index)
+    LargeBits GetFullProof(const uint8_t* challenge, uint32_t index, bool disable_parallel_reads = false)
```
xdustinface, line 204:
```suggestion
    LargeBits GetFullProof(const uint8_t* challenge, uint32_t index, const bool disable_parallel_reads = false)
```

src/prover_disk.hpp
```diff
@@ -635,11 +639,18 @@ class DiskProver {
...
-        // Create individual file handles to allow parallel processing
-        std::ifstream disk_file(filename, std::ios::in | std::ios::binary);
-        uint128_t line_point = ReadLinePoint(disk_file, depth, position);
+        uint128_t line_point;
+
+        if (!disk_file) {
+            // No disk file passed in, so we assume here we are doing parallel reads
+            // Create individual file handles to allow parallel processing
+            std::ifstream disk_file(filename, std::ios::in | std::ios::binary);
+            line_point = ReadLinePoint(disk_file, depth, position);
```
xdustinface, line 650:
```suggestion
            std::ifstream disk_file_parallel(filename, std::ios::in | std::ios::binary);
            line_point = ReadLinePoint(disk_file_parallel, depth, position);
```

To not shadow the parameter

xdustinface (changes requested):
See some comments below + one question: What do you think about removing this weird "double bool logic" by renaming all `disable_parallel_read` to `parallel_reads` which is then defaulted to be on and of course adjust the related descriptions? Same for the `chia-blockchain` PR.

src/prover_disk.hpp
emlowe: I'm not opposed to this - initially I had it the other way . Personally, I find it slightly more obvious this way that you are disabling the feature. I assume you would want this then also in the config.yaml as "parallel_reads: false".

src/cli.cpp
emlowe: I did this on purpose. Do you want this changed to `--parallel_read=false` to turn off?

src/prover_disk.hpp
emlowe: const doesn't do much IMO with pass-by-value parameters

src/prover_disk.hpp
xdustinface: This particular diff doesn't change any behavior, its just "be explicit in the condition" i.e. `!x` has the same outcome as `x == nullptr`

> I'm not opposed to this - initially I had it the other way . Personally, I find it slightly more obvious this way that you are disabling the feature.

I guess you are talking about the renaming of the flag here? If so:

> I assume you would want this then also in the config.yaml as "parallel_reads: false".

Yeah :+1: that's what i meant with: 

> Same for the `chia-blockchain` PR.

arvidn:
Can you link to the related tickets/bug reports?

emlowe:
In conjunction with the new config.yaml option - this should address https://github.com/Chia-Network/chia-blockchain/issues/7603 and https://github.com/Chia-Network/chia-blockchain/issues/7359

emlowe:
I'll note the following here:
Switching to use std::filesystem from gulrak on macos did not affect the behaviour (remains slow)
I tested with FAT32 and was not able to reproduce (FAT32 does not seem affected)
Testing was primarily done with timing runs of chia plots check and k=25 plots
Adding an option to revert to the old behaviour seemed easier than trying to diagnose why the parallel code is so slow in this one particular case (although it is fairly peculiar)

fchirica (approved):
lgtm

PR #297, opened 2021-07-21
El.fix ci runners

emlowe:
Make the C++ test into separate jobs so they run in parallel - this speeds up the test completion
Make sure to increase swap space to workaround as yet undetermined high memory usage during ASAN and TSAN tests


emlowe:
opps, messed up this PR - closing

PR #299, opened 2021-07-26
Removes redundant booleans

joaquimguimaraes:
After the if clause (shift >= 128) fails, it is ensured that shift is less than 128 when it goes through that part of the code, hence (128 > shift) is the same as (shift < 128) which is always true. Therefore it can be removed from both clauses.


arvidn:
this change is correct and silences the LGTM issues. I looked at this months ago, and I have some issues with this change:

1. I don't believe there are any tests covering this change
2. There may be tests in the original repository that we copy-pasted this code from, but we're not running them
3. We really should use the version from the original repo and contribute changes like these (and perhaps tests if there aren't any)

PR #302, opened 2021-08-05
src|python-bindings: Use `std::vector<uint8_t>` for `id` and `memo`

xdustinface:


src/prover_disk.hpp
```diff
@@ -80,16 +80,14 @@ class DiskProver {
...
+        memcpy(id.data(), header.id, sizeof(header.id));
         this->k = header.k;
         SafeSeek(disk_file, offsetof(struct plot_header, fmt_desc) + fmt_desc_len);
 
         uint8_t size_buf[2];
         SafeRead(disk_file, size_buf, 2);
-        this->memo_size = Util::TwoBytesToInt(size_buf);
-        this->memo = new uint8_t[this->memo_size];
-        SafeRead(disk_file, this->memo, this->memo_size);
+        memo.resize(Util::TwoBytesToInt(size_buf));
```
arvidn, line 89: this function should have been called `TwoBytesToUint()`. it does return an unsigned int so this is safe, but I had to look it up.

src/prover_disk.hpp
```diff
@@ -122,18 +120,15 @@ class DiskProver {
...
         for (int i = 0; i < 6; i++) {
             Encoding::ANSFree(kRValues[i]);
         }
         Encoding::ANSFree(kC3R);
     }
 
-    void GetMemo(uint8_t* buffer) { memcpy(buffer, memo, this->memo_size); }
-
-    uint32_t GetMemoSize() const noexcept { return this->memo_size; }
+    std::vector<uint8_t> GetMemo() { return memo; }
```
arvidn, line 129: would it make sense for this to return a `span<const uint8_t>` or a `const std::vector<uint8_t>&`? (as well as `GetId()` below)

src/prover_disk.hpp
```diff
@@ -577,7 +571,7 @@ class DiskProver {
     //     Where a < b is defined as:  max(b) > max(a) where a and b are lists of k bit elements
     std::vector<LargeBits> ReorderProof(const std::vector<Bits>& xs_input) const
     {
-        F1Calculator f1(k, id);
+        F1Calculator f1(k, id.data());
```
arvidn, line 574: heh, we're hoping that `id` is large enough here :) (it would be nice for the `F1Calculator to take a span in the future)

arvidn (approved):

src/prover_disk.hpp
xdustinface: Wasn't me! :smile:

src/prover_disk.hpp
xdustinface: I agree `span` would be nice in many places but we have C++17 set here still so this feels out of scope. I changed the return to const reference though :+1:

src/prover_disk.hpp
xdustinface: We had the same hope before this PR but i see your point.

python-bindings/chiapos.cpp
```diff
@@ -80,20 +80,14 @@ PYBIND11_MODULE(chiapos, m)
         .def(
             "get_memo",
             [](DiskProver &dp) {
-                uint8_t *memo = new uint8_t[dp.GetMemoSize()];
-                dp.GetMemo(memo);
-                py::bytes ret = py::bytes(reinterpret_cast<char *>(memo), dp.GetMemoSize());
-                delete[] memo;
-                return ret;
+                std::vector<uint8_t> memo = dp.GetMemo();
+                return py::bytes(reinterpret_cast<char*>(memo.data()), memo.size());
```
arvidn, line 84: Its probably not a big deal, but this will copy the buffer twice, where i would think once is enough

python-bindings/chiapos.cpp
```diff
@@ -80,20 +80,14 @@ PYBIND11_MODULE(chiapos, m)
...
             })
         .def(
             "get_id",
             [](DiskProver &dp) {
-                uint8_t *id = new uint8_t[kIdLen];
-                dp.GetId(id);
-                py::bytes ret = py::bytes(reinterpret_cast<char *>(id), kIdLen);
-                delete[] id;
-                return ret;
+                std::vector<uint8_t> id = dp.GetId();
```
arvidn, line 89: Same here

arvidn (approved):
This is a clear improvement

python-bindings/chiapos.cpp
xdustinface: Changed!

PR #303, opened 2021-08-06
python-bindings|prover_disk|tests: Add `pickle` support for `DiskProver`

xdustinface:
This is needed for plots caching in `chia-blockchain`.

Based on #302

### DIFF
--- a/python-bindings/chiapos.cpp
+++ b/python-bindings/chiapos.cpp
@@ -78,6 +78,28 @@ PYBIND11_MODULE(chiapos, m)
     py::class_<DiskProver>(m, "DiskProver")
         .def(py::init<const std::string &>())
         .def("is_valid", [](const DiskProver &dp) { return dp.IsValid(); })
+        .def(py::pickle(
+            [](const DiskProver& dp) { // __getstate__
+                return py::make_tuple(dp.GetFilename(),
+                                      dp.GetSize(),
+                                      dp.GetMemo(),
+                                      dp.GetId(),
+                                      dp.GetTableBeginPointers(),
+                                      dp.GetC2());
+                },
+            [](const py::tuple& t) { // __setstate__
+                if (t.size() != 6)
+                    throw std::runtime_error("Invalid state!");
+
+                auto filename = t[0].cast<std::string>();
+                auto k = t[1].cast<uint8_t>();
+                auto memo = t[2].cast<std::vector<uint8_t>>();
+                auto id = t[3].cast<std::vector<uint8_t>>();
+                auto table_begin_pointers = t[4].cast<std::vector<uint64_t>>();
+                auto C2 = t[5].cast<std::vector<uint64_t>>();
+                return DiskProver(filename, memo, id, k,
+                                  table_begin_pointers, C2);
+            }))
         .def(
             "get_memo",
             [](DiskProver &dp) {
--- a/python-bindings/chiapos.cpp
+++ b/python-bindings/chiapos.cpp
@@ -77,6 +77,7 @@ PYBIND11_MODULE(chiapos, m)
 
     py::class_<DiskProver>(m, "DiskProver")
         .def(py::init<const std::string &>())
+        .def("is_valid", [](const DiskProver &dp) { return dp.IsValid(); })
         .def(
             "get_memo",
             [](DiskProver &dp) {
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -78,7 +78,7 @@ public:
         if (this->table_begin_pointers.size() != table_begin_pointers_size) {
             throw std::invalid_argument("Invalid table_begin_pointers size: " + std::to_string(this->table_begin_pointers.size()));
         }
-        uint32_t expected_size = ((this->table_begin_pointers[10] - this->table_begin_pointers[9]) / (Util::ByteAlign(k) / 8)) - 1;
+        const uint32_t expected_size = ((this->table_begin_pointers[10] - this->table_begin_pointers[9]) / (Util::ByteAlign(k) / 8)) - 1;
         if (this->C2.size() != expected_size) {
             throw std::invalid_argument("Invalid C2 size: " + std::to_string(this->C2.size()));
         }
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -47,7 +47,7 @@ struct plot_header {
 // The DiskProver, given a correctly formatted plot file, can efficiently generate valid proofs
 // of space, for a given challenge.
 class DiskProver {
-    const size_t table_begin_pointers_size{11};
+    static const size_t table_begin_pointers_size{11};
 public:
     // The constructor opens the file, and reads the contents of the file header. The table pointers
     // will be used to find and seek to all seven tables, at the time of proving.
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -47,7 +47,7 @@ struct plot_header {
 // The DiskProver, given a correctly formatted plot file, can efficiently generate valid proofs
 // of space, for a given challenge.
 class DiskProver {
-    const size_t nTableBeginPointerCount{11};
+    const size_t table_begin_pointers_size{11};
 public:
     // The constructor opens the file, and reads the contents of the file header. The table pointers
     // will be used to find and seek to all seven tables, at the time of proving.
@@ -75,11 +75,11 @@ public:
         if (k < kMinPlotSize || k > kMaxPlotSize) {
             throw std::invalid_argument("Invalid k: " + std::to_string(k));
         }
-        if (this->table_begin_pointers.size() != nTableBeginPointerCount) {
+        if (this->table_begin_pointers.size() != table_begin_pointers_size) {
             throw std::invalid_argument("Invalid table_begin_pointers size: " + std::to_string(this->table_begin_pointers.size()));
         }
-        uint32_t nExpectedSize = ((this->table_begin_pointers[10] - this->table_begin_pointers[9]) / (Util::ByteAlign(k) / 8)) - 1;
-        if (this->C2.size() != nExpectedSize) {
+        uint32_t expected_size = ((this->table_begin_pointers[10] - this->table_begin_pointers[9]) / (Util::ByteAlign(k) / 8)) - 1;
+        if (this->C2.size() != expected_size) {
             throw std::invalid_argument("Invalid C2 size: " + std::to_string(this->C2.size()));
         }
     }
@@ -297,11 +297,11 @@ private:
         memo_in.resize(Util::TwoBytesToInt(size_buf));
         SafeRead(disk_file, memo_in.data(), memo_in.size());
 
-        table_begin_pointers_in = std::vector<uint64_t>(nTableBeginPointerCount, 0);
+        table_begin_pointers_in = std::vector<uint64_t>(table_begin_pointers_size, 0);
         C2_in = std::vector<uint64_t>();
 
         uint8_t pointer_buf[8];
-        for (size_t i = 1; i < nTableBeginPointerCount; i++) {
+        for (size_t i = 1; i < table_begin_pointers_size; i++) {
             SafeRead(disk_file, pointer_buf, 8);
             table_begin_pointers_in[i] = Util::EightBytesToInt(pointer_buf);
         }
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -53,73 +53,13 @@ public:
     // will be used to find and seek to all seven tables, at the time of proving.
     explicit DiskProver(const std::string& filename) : id(kIdLen)
     {
-        struct plot_header header{};
         this->filename = filename;
-
-        std::ifstream disk_file(filename, std::ios::in | std::ios::binary);
-
-        if (!disk_file.is_open()) {
-            throw std::invalid_argument("Invalid file " + filename);
-        }
-        // 19 bytes  - "Proof of Space Plot" (utf-8)
-        // 32 bytes  - unique plot id
-        // 1 byte    - k
-        // 2 bytes   - format description length
-        // x bytes   - format description
-        // 2 bytes   - memo length
-        // x bytes   - memo
-
-        SafeRead(disk_file, (uint8_t*)&header, sizeof(header));
-        if (memcmp(header.magic, "Proof of Space Plot", sizeof(header.magic)) != 0)
-            throw std::invalid_argument("Invalid plot header magic");
-
-        uint16_t fmt_desc_len = Util::TwoBytesToInt(header.fmt_desc_len);
-
-        if (fmt_desc_len == kFormatDescription.size() &&
-            !memcmp(header.fmt_desc, kFormatDescription.c_str(), fmt_desc_len)) {
-            // OK
-        } else {
-            throw std::invalid_argument("Invalid plot file format");
-        }
-        memcpy(id.data(), header.id, sizeof(header.id));
-        this->k = header.k;
-        SafeSeek(disk_file, offsetof(struct plot_header, fmt_desc) + fmt_desc_len);
-
-        uint8_t size_buf[2];
-        SafeRead(disk_file, size_buf, 2);
-        memo.resize(Util::TwoBytesToInt(size_buf));
-        SafeRead(disk_file, memo.data(), memo.size());
-
-        this->table_begin_pointers = std::vector<uint64_t>(nTableBeginPointerCount, 0);
-        this->C2 = std::vector<uint64_t>();
-
-        uint8_t pointer_buf[8];
-        for (size_t i = 1; i < nTableBeginPointerCount; i++) {
-            SafeRead(disk_file, pointer_buf, 8);
-            this->table_begin_pointers[i] = Util::EightBytesToInt(pointer_buf);
-        }
-
-        SafeSeek(disk_file, table_begin_pointers[9]);
-
-        uint8_t c2_size = (Util::ByteAlign(k) / 8);
-        uint32_t c2_entries = (table_begin_pointers[10] - table_begin_pointers[9]) / c2_size;
-        if (c2_entries == 0 || c2_entries == 1) {
-            throw std::invalid_argument("Invalid C2 table size");
-        }
-
-        // The list of C2 entries is small enough to keep in memory. When proving, we can
-        // read from disk the C1 and C3 entries.
-        auto* c2_buf = new uint8_t[c2_size];
-        for (uint32_t i = 0; i < c2_entries - 1; i++) {
-            SafeRead(disk_file, c2_buf, c2_size);
-            this->C2.push_back(Bits(c2_buf, c2_size, c2_size * 8).Slice(0, k).GetValue());
-        }
-
-        delete[] c2_buf;
+        Read(id, memo, k, table_begin_pointers, C2);
     }
 
     // Note: This constructor presumes the input parameter are valid for the provided file and does
-    // not validate the file itself.
+    // not validate the file itself. It's recommended to use `IsValid` after construction to make
+    // sure the given file matches expectations.
     DiskProver(std::string filename, std::vector<uint8_t> memo, std::vector<uint8_t> id, uint8_t k,
                std::vector<uint64_t> table_begin_pointers, std::vector<uint64_t> C2) :
         filename(std::move(filename)),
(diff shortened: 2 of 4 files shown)


python-bindings/chiapos.cpp
```diff
@@ -77,23 +77,39 @@ PYBIND11_MODULE(chiapos, m)
...
+                },
+                [](const py::tuple& t) { // __setstate__
+                if (t.size() != 6)
+                    throw std::runtime_error("Invalid state!");
+
+                auto filename = t[0].cast<std::string>();
+                auto k = t[1].cast<uint8_t>();
+                auto memo = t[2].cast<std::vector<uint8_t>>();
+                auto id = t[3].cast<std::vector<uint8_t>>();
+                auto table_begin_pointers = t[4].cast<std::vector<uint64_t>>();
```
arvidn, line 97: I have a feeling that these are offsets into the plot file, and that we should check that they are indeed all smaller than the size of the plot file. We've had many problems where we crash on corrupt or truncated plot files in the past.

python-bindings/chiapos.cpp
```diff
@@ -77,23 +77,39 @@ PYBIND11_MODULE(chiapos, m)
...
+                                      dp.GetC2());
+                },
+                [](const py::tuple& t) { // __setstate__
+                if (t.size() != 6)
+                    throw std::runtime_error("Invalid state!");
+
+                auto filename = t[0].cast<std::string>();
+                auto k = t[1].cast<uint8_t>();
+                auto memo = t[2].cast<std::vector<uint8_t>>();
+                auto id = t[3].cast<std::vector<uint8_t>>();
```
arvidn, line 96: The memo and Id are required to be 32 bytes, right? I'm not sure about the length, but we should probably verify that they are valid

python-bindings/chiapos.cpp
```diff
@@ -77,23 +77,39 @@ PYBIND11_MODULE(chiapos, m)
...
+            }))
         .def(
             "get_memo",
             [](DiskProver &dp) {
-                uint8_t *memo = new uint8_t[dp.GetMemoSize()];
-                dp.GetMemo(memo);
-                py::bytes ret = py::bytes(reinterpret_cast<char *>(memo), dp.GetMemoSize());
-                delete[] memo;
-                return ret;
+                std::vector<uint8_t> memo = dp.GetMemo();
```
arvidn, line 105: +1 for fixing the manual memory management

src/prover_disk.hpp
```diff
@@ -119,21 +117,43 @@ class DiskProver {
...
+
+    DiskProver(DiskProver&& other) :
+        filename(std::move(other.filename)),
+        memo(std::move(other.memo)),
+        id(std::move(other.id)),
+        k(other.k),
+        table_begin_pointers(std::move(other.table_begin_pointers)),
+        C2(std::move(other.C2))
+    {
+    }
```
arvidn, line 139: It looks like you could just say = default here. Does your move constructor differ in some way? If so, that should probably be called out in a comment

src/prover_disk.hpp
```diff
@@ -119,21 +117,43 @@ class DiskProver {
...
 
+    DiskProver(std::string filename, std::vector<uint8_t> memo, std::vector<uint8_t> id, uint8_t k,
+               std::vector<uint64_t> table_begin_pointers, std::vector<uint64_t> C2) :
+        filename(std::move(filename)),
+        memo(std::move(memo)),
+        id(std::move(id)),
+        k(k),
+        table_begin_pointers(std::move(table_begin_pointers)),
+        C2(std::move(C2))
+    {
```
arvidn, line 128: Maybe this would be a good place to validate the inputs

arvidn (commented):
I think the new constructor should validate its inputs, and have tests ensuring it rejects invalid inputs correctly. (I.e. we shouldn't trust the pickled file to be uncorrupted or trustworthy)

src/prover_disk.hpp
xdustinface: Default doesn't work here because `std::mutex` has the move constructor deleted.

arvidn: so, does the moved from object need to have its mutex locked? what does the mutex protect exactly? This sounds like it would need a bit more comprehensive description (in a comment)

xdustinface: Good point, locked the mutex while moving but i think the move constructor itself does not really need a comment section about what the mutex does this should be part of the mutex definition imo and is a bit out of scope of this PR.

arvidn:
I think it's a bit weird to lock the mutex in the move constructor, just like in the destructor. Another thread better not have a reference to the object as it's about to be destructed. If it does, locking the murex won't help.

Some of the getters seems to copy the vectors a bit gratuitously as well.

python-bindings/chiapos.cpp
xdustinface: I was thinking about this now for a while and i was even about to just us the normal filename constructor like in [here](https://github.com/Chia-Network/chiapos/compare/18136f09ec1c55d5be3f418b1d244ecc0d7d3f37..438ff5e98747a5399f81be80639a9f91d9bf2878) but now after i just pushed the change it came to my mind that this will cause trouble if we depend on the file while de-serializing if the file does not longer exist.

arvidn:
My comment in the G1 pickle support applies here too. Can we rely on pickle being a stable format across python versions?

We need to ensure we don't accidentally break backwards compatibility.

xdustinface:
> Another thread better not have a reference to the object as it's about to be destructed.

Hmm better not yeah but how to prevent it? If we don't lock it in the move constructor we could end up moving the object while another thread has the mutex locked i.e. its still fetching a proof or so, no?

> My comment in the G1 pickle support applies here too. Can we rely on pickle being a stable format across python versions?
> 
> We need to ensure we don't accidentally break backwards compatibility.

The soon following change in `chia-blockchain` will cache the following data structure:

```
@dataclass
class Cache:
    version: int
    plots: Dict[Path, PlotInfo]
    plot_filename_paths: Dict[str, Tuple[str, Set[str]]]
    failed_to_open_filenames: Dict[Path, int]
    no_key_filenames: Set[Path]
```

which contains the `version` field i guess that's good enough?

python-bindings/chiapos.cpp
arvidn: What *should* happen if the file doesn't exist? Today we don't populate the plot list with the file to begin with. Throwing an exception, I suppose, would stop loading of all other plots...

src/prover_disk.hpp
arvidn: What I meant is that; looking at that code, it looks like a mistake to not just do = default, and people may be tempted to make that change in the future. A comment would justify why it *isn't* just = default

arvidn:
> > Another thread better not have a reference to the object as it's about to be destructed.
> 
> Hmm better not yeah but how to prevent it? If we don't lock it in the move constructor we could end up moving the object while another thread has the mutex locked i.e. its still fetching a proof or so (relies on the data we are about to move), no?

My point is that locking the mutex won't help in that case. We have UB either way. So, we could assert that the mutex isn't locked maybe, but fundamentally, I think we just have to assume the users of this class aren't doing shady stuff with it.

> 
> > We need to ensure we don't accidentally break backwards compatibility.
> 
> which contains the `version` field i guess that's good enough?

If the pickle file format changes, you won't ever get to see the version number.

Using our Streamable format may be an alternative, but that would create a dependency on the chia-blockchain repo which would be a hassle.

python-bindings/chiapos.cpp
xdustinface: Well, it shouldn't fail to de-serialize the whole cached data if only one plot file is not longer available but this will happen if we throw in the constructor because the file doesn't exist. The idea was to just serialize the whole list of plots and load it on startup like in https://github.com/xdustinface/chia-blockchain/commit/2ca417ab548e9a34fec43b27c1ed3ca0009259d0. The `PlotManager` would then take care of removed plots in is refresh task. With your proposed change we would need to store a cache file for each plot.

arvidn: I'm only proposing this case is handled well (and maybe even tested). But also the case where the pickled file is corrupt or the plot is corrupt.

xdustinface:
> My point is that locking the mutex won't help in that case. We have UB either way. So, we could assert that the mutex isn't locked maybe, but fundamentally, I think we just have to assume the users of this class aren't doing shady stuff with it.

I mean, technically there is still a possible case where its no UB if we lock the mutex like if we have two threads and we trigger stop for both, one of them has currently the mutex locked and the other wants to move the object to a shutdown thread or whatever idea one might have, right?  But if we don't lock the mutex this case would be an UB. Not that this is super likely but its there and a reason to just lock it unless i miss something.

> If the pickle file format changes, you won't ever get to see the version number.

You can still always use older protocols if there will be a non-backwards compatible protocol bump at some point.

```
https://docs.python.org/3/library/pickle.html#module-pickle
....
The pickle serialization format is guaranteed to be backwards compatible across Python releases provided a compatible pickle protocol is chosen and pickling and unpickling code deals with Python 2 to Python 3 type differences if your data is crossing that unique breaking change language boundary.
....
```

> Using our Streamable format may be an alternative, but that would create a dependency on the chia-blockchain repo which would be a hassle.

Thats what i started with... but not to add `Streamable` as dependency here but rather just to implement a `__bytes__` and a `from_bytes` for `DiskProver` in the python bindings by adding serialization to the C++ part and have `DiskProver` serialized as `Streamable` in `chia-blockchain` then. But this also required to change a lot stuff in the `Streamable` implementation because it doesn't support `Dict`, `Set` and the dynamic custom types at this point. I got the custom types working kind of by prepending the size of the serialized data in the pybind implementation before returning in `__bytes__` but for the `Dict` part i ended up in a blind end at some point where i wasn't able to de-serialize it because of some weird "`isinstance` can't be used for nested type annotations" thing. I even already spent quite some time implementing basic serialization for the C++ part here https://github.com/xdustinface/chiapos/commit/a84ad93c10323a94d7e9650600e42cea765da41c and for `DiskProver` itself here https://github.com/xdustinface/chiapos/commit/748c3df7a7de62486568e98cfb875e085865bbf6. Thats what i also wanted to use for the implementation of the pickle support initially but basically the moment i wanted to submit the PR for the serialization part i realized its simpler to just add another constructor for `DiskProver` to make it simpler and add less new code :smile:

src/prover_disk.hpp
xdustinface: Ahh, i see, but if the mutex gets locked in the move constructor its pretty obvious without a comment why its not just default, isn't it? I'm going to wait for your reply in the outer scope (we should have only talked about it in this subsection haha) :) I can add that comment up without the lock.

python-bindings/chiapos.cpp
xdustinface: I'm don't get what you want/mean here with the corrupt pickle/plot file as this only adds support for pickle. Last push added basic validation and a comment to point out that the new constructor assumes the parameter match the underlying file. Any thoughts/objections?

arvidn: is it possible to violate the class invariants via the constructor you use when un-pickling?

xdustinface: Obviously, like the comment says you can provide parameter that doesn't match the underlying file but so can you load a corrupt plot without knowing with the other constructor, right? I could of course add some `void EnsureValidated()` method which checks a `fIsValidated` flag before each read to verify the file pointers are in bounds if that flag is false but at the and this doesn't improve the constructor and we use `SafeRead` everywhere which *should* throw if something doesn't match and all what may happen if the pickled file for some weird reason i can't think about gets altered in a way that a pointer exceeds the bounds nothing bad can happen besides a failing lookup or do i miss something here?

arvidn: > Obviously, like the comment says you can provide parameter that doesn't match the underlying file but so can you load a corrupt plot without knowing with the other constructor, right?

I don't know. I'm not very familiar with this class. It sounds like it doesn't have very many invariants then.

> all what may happen if the pickled file for some weird reason i can't think about gets altered in a way that a pointer exceeds the bounds nothing bad can happen besides a failing lookup or do i miss something here?

It's hard to know without a test.

xdustinface: The last two commits and this https://github.com/Chia-Network/chia-blockchain/pull/8009/files#diff-1c950ad1eb2a79b51cb4f61cf39b8fe8d09d28ad963ce7dd24107053ff316d14R113-R126 should take care of this. Any objections or what else do you want to see tested related to this?

arvidn:
I think the `EnsureValid()` patch should be reverted.

Generally, classes should avoid using 2-phase initialization. If a constructor fails to establish the class invariant, it should fail. That way, all instances of a class are known good, no half-initialized objects.

It seems the reason why you want the `DiskProver` to have such "invalid" or half-initialized state is because the de-serialization code (de-pickling) isn't controlled by you, and it doesn't support the concept of individual items in a list failing, without failing everything.

The modern alternative to 2-phase initialization is to use `std::optional<DiskProver>`. If all of that is too impractical, I think the *impact* of the compromise to use 2-phase initialization should at least be mitigated and isolated. The `EnsureValid()` patch exposes the possibly-uninitialized nature of this class to all of its users, that's why I think it should be reverted.

exposing `is_valid()` to allow filtering out objects that failed to load immediately after de-pickling seems like a reasonable compromise.

xdustinface:
@arvidn I restructured the commit history a bit and put some stuff into separate commits. I just saw windows tests failing for some reason but i need to sleep now.. will check that when i wake up.

> The modern alternative to 2-phase initialization is to use `std::optional<DiskProver>`. If all of that is too impractical, I think the _impact_ of the compromise to use 2-phase initialization should at least be mitigated and isolated. The `EnsureValid()` patch exposes the possibly-uninitialized nature of this class to all of its users, that's why I think it should be reverted.

Unfortunately it seems like its not possible to use `std::optional` with pybind/pickle here. I tried a bit but couldn't get it working in reasonable time. If you have a solution feel free to provide a commit. Otherwise, see below.

> I think the `EnsureValid()` patch should be reverted.

I refactored it, see b395860. Note that the alternative constructor still throws the same way like it did before but the new `DiskProver::Validate` is not longer enforced by all class methods, it just throws if it gets called + there is still the non-throwing `DiskProver::IsValid`.

src/prover_disk.hpp
```diff
@@ -47,74 +47,52 @@ struct plot_header {
 // The DiskProver, given a correctly formatted plot file, can efficiently generate valid proofs
 // of space, for a given challenge.
 class DiskProver {
+    const size_t nTableBeginPointerCount{11};
```
arvidn, line 50: I don't think it's a good idea to add an immutable member like this. This makes `DiskProver` objects not movable.
Did you mean to make this `static`?
Also, I don't think we should use hungarian notation.

src/prover_disk.hpp
```diff
@@ -47,74 +47,52 @@ struct plot_header {
...
+            throw std::invalid_argument("Invalid table_begin_pointers size: " + std::to_string(this->table_begin_pointers.size()));
         }
-
-        // The list of C2 entries is small enough to keep in memory. When proving, we can
-        // read from disk the C1 and C3 entries.
-        auto* c2_buf = new uint8_t[c2_size];
-        for (uint32_t i = 0; i < c2_entries - 1; i++) {
-            SafeRead(disk_file, c2_buf, c2_size);
-            this->C2.push_back(Bits(c2_buf, c2_size, c2_size * 8).Slice(0, k).GetValue());
+        uint32_t nExpectedSize = ((this->table_begin_pointers[10] - this->table_begin_pointers[9]) / (Util::ByteAlign(k) / 8)) - 1;
```
arvidn, line 81: this should be `const`

src/prover_disk.hpp
```diff
@@ -47,74 +47,52 @@ struct plot_header {
...
-
-        uint16_t fmt_desc_len = Util::TwoBytesToInt(header.fmt_desc_len);
-
-        if (fmt_desc_len == kFormatDescription.size() &&
-            !memcmp(header.fmt_desc, kFormatDescription.c_str(), fmt_desc_len)) {
-            // OK
-        } else {
-            throw std::invalid_argument("Invalid plot file format");
+        if (k < kMinPlotSize || k > kMaxPlotSize) {
+            throw std::invalid_argument("Invalid k: " + std::to_string(k));
```
arvidn, line 76: the comment above this constructor suggests the inputs are not validated, but the seem to be, at least to some extent. Are there some aspects left that aren't validated, or is the comment outdated?

src/prover_disk.hpp
```diff
@@ -238,10 +253,78 @@ class DiskProver {
     std::string filename;
     std::vector<uint8_t> memo;
     std::vector<uint8_t> id;  // Unique plot id
-    uint8_t k;
+    uint8_t k{0};
     std::vector<uint64_t> table_begin_pointers;
     std::vector<uint64_t> C2;
 
+    void Read(std::vector<uint8_t>& id_in, std::vector<uint8_t>& memo_in, uint8_t& k_in, std::vector<uint64_t>& table_begin_pointers_in, std::vector<uint64_t>& C2_in) const
```
arvidn, line 260: the names of these parameters confuse me. The `_in` suffix suggests they are *in*parameters, but they all appear to be output parameters. Did you consider returning them in a tuple, as multiple return values?

src/prover_disk.hpp
```diff
@@ -238,10 +253,78 @@ class DiskProver {
...
+        k_in = header.k;
+        SafeSeek(disk_file, offsetof(struct plot_header, fmt_desc) + fmt_desc_len);
+
+        uint8_t size_buf[2];
+        SafeRead(disk_file, size_buf, 2);
+        memo_in.resize(Util::TwoBytesToInt(size_buf));
+        SafeRead(disk_file, memo_in.data(), memo_in.size());
+
+        table_begin_pointers_in = std::vector<uint64_t>(nTableBeginPointerCount, 0);
+        C2_in = std::vector<uint64_t>();
```
arvidn, line 301: The only benefit I can think of, of passing these vectors in as out-parameters is that they can reuse memory they may already have allocated. However, when you assign them like this, rather than `.resize(0)`, you defeat that on benefit.

I would think, though, that these allocations are cheap anyway, and that it would make sense to return them instead of taking them as out-parameters.

src/prover_disk.hpp
```diff
@@ -238,10 +253,78 @@ class DiskProver {
...
     std::vector<uint8_t> memo;
     std::vector<uint8_t> id;  // Unique plot id
-    uint8_t k;
+    uint8_t k{0};
     std::vector<uint64_t> table_begin_pointers;
     std::vector<uint64_t> C2;
 
+    void Read(std::vector<uint8_t>& id_in, std::vector<uint8_t>& memo_in, uint8_t& k_in, std::vector<uint64_t>& table_begin_pointers_in, std::vector<uint64_t>& C2_in) const
+    {
+        std::lock_guard<std::mutex> lock(_mtx);
```
arvidn, line 262: the mutex didn't used to be locked when reading this, as far as I can tell. This is only called from the filename constructor, right? I see it's also called from `Validate()`. It looks like this mutex is unneccessary though, but we can leave that for a future patch.

src/prover_disk.hpp
```diff
@@ -238,10 +253,78 @@ class DiskProver {
...
+
+        // The list of C2 entries is small enough to keep in memory. When proving, we can
+        // read from disk the C1 and C3 entries.
+        auto* c2_buf = new uint8_t[c2_size];
+        for (uint32_t i = 0; i < c2_entries - 1; i++) {
+            SafeRead(disk_file, c2_buf, c2_size);
+            C2_in.push_back(Bits(c2_buf, c2_size, c2_size * 8).Slice(0, k).GetValue());
+        }
+
+        delete[] c2_buf;
```
arvidn, line 325: this leaks if `SafeRead()` throws an exception. This was an issue before your patch, I just want to make a note to remember to fix it later

arvidn (commented):
It's not clear to me how or why one would use the `Validate()` or `is_valid()` functions. It would seem simpler to just attempt to look up a proof and have it fail. It's fine to defer the failure until we actually need to do the lookup, right?

src/prover_disk.hpp
xdustinface: > This makes `DiskProver` objects not movable.

What do you mean? Maybe not by default but its still movable :man_shrugging: However, i agree, it makes sense to have it `static`. 

> Also, I don't think we should use hungarian notation.

I dropped it in 96f0445 to stay consistent with the codebase, but whats your point here? I'm just curious because i'm used to always use it for primitives or standard container types.

src/prover_disk.hpp
xdustinface: See 94cad1af9eeebe6d2b5a0e80be3421b1aa6fd3ed

src/prover_disk.hpp
xdustinface: It does not say inputs are not validated. It says the inputs are not validated against the file.

src/prover_disk.hpp
xdustinface: I feel like a tuple would make this more confusing than better. See a9ca08363e61ac45ed9c3ccae7840464818a7344.

src/prover_disk.hpp
xdustinface: I didn't use the `resize` approach here because i wanted to save one line :smile: (setting the contents or at least the first element of `table_begin_pointers_out` to zero) and it obviously doesn't really matter given the size of this vectors.

xdustinface:
> It's not clear to me how or why one would use the `Validate()` or `is_valid()` functions. It would seem simpler to just attempt to look up a proof and have it fail. It's fine to defer the failure until we actually need to do the lookup, right?

I'm using it to drop invalid provers after cache loading, see https://github.com/Chia-Network/chia-blockchain/pull/8009/commits/3b0b971df8da225c13cf99be5b56ce03b82f437f#diff-1c950ad1eb2a79b51cb4f61cf39b8fe8d09d28ad963ce7dd24107053ff316d14R116

src/prover_disk.hpp
arvidn: I believe there is widespread agreement that encoding type information in variable names is undesirable, as documented in the [cpp core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-name-type).

arvidn: > What do you mean? Maybe not by default but its still movable

Right, I noticed later that you implement the move constructor explicitly, and just don't touch this member.

src/prover_disk.hpp
arvidn: ok

arvidn:
> I'm using it to drop invalid provers after cache loading,

It would seem to me that doing that would defeat the purpose of caching these entries in the first place. If we're reading it back from the files unconditionally, to validate it, couldn't we just do that to begin with, and not also read the pickled cache?

Perhaps I'm missing something

arvidn (approved):

xdustinface:
> is this PR abandoned now?

Yes!

PR #306, opened 2021-09-03
Add buffer size checks that fail if encoded_size is greater than the …

prozacchiwawa:
…amount read from disk, c3_entry_size - 2

Admittedly I'm not familiar with this code, and this is a deduction from looking at the crash https://github.com/Chia-Network/chia-blockchain/issues/8303.  My hypothesis is that the crash, which occurs at a movzbl instruction which i believe is reading a byte from cSrc in FSE_decompress_usingDTable_generic.


arvidn:
I think one great addition to this would be tests that constructs invalid plots to trigger these conditions. maybe that's difficult to create, but I would imagine it would be deterministic at least

Issue #313, opened 2021-10-20
Rename rather than copy if tmp2 filename is on the same device as final filename


cross:
Drawing this over from Chia-Network/chia-blockchain#8847 -

When tmp and final files are in different directories on the same device, the code doesn't know that it can perform a rename(), and instead spends a lot of time unnecessarily copying the file data.

cross:
Not stale, waiting on #314

cross:
Not stale, waiting on #314

cross:
This is an ongoing irritation in my environment.  Can I help in any way getting #314 testing completed so it can be merged?

cross:
Not stale, still need #314 merged.

cross:
Still waiting.  @hoffmang9 , can someone get this in please?

wjblanke:
We'll put this into 1.3 beta so we can get a better idea of how it works in practice with different drives, devices, os, mount point, etc. Need to do another chiapos release then merge into main

PR #314, opened 2021-10-20
If tmp2 and final files are in same filesystem, rename() not copy()

cross:
This avoids potentially painful copies of large files within the same filesystem.
To note:
 a) I'm a UNIX person, so I'm not sure if this will compile on Windows as coded.  Directions welcome.
 b) I didn't write a test for this, it appears all of the existing test only use files in the same directory.  Is the copy codepath in CreatePlotDisk() tested somewhere?


cross:
This fixes #313 , waiting on workflows to test windows applicability.

cross:
Can someone merge this?  Is anything else needed?

hoffmang9:
This is likely to take a bit broader testing and that's the only hesitation. Right this moment there are other things at a higher priority. Should it test well on BSD/MacOS both flavors, three FSs etc.

cross:
> This is likely to take a bit broader testing and that's the only hesitation. Right this moment there are other things at a higher priority. Should it test well on BSD/MacOS both flavors, three FSs etc.

Okay.  It's very old libc calls, so I'm sure all UNIX/Linux variants and Mac OS (>9) will work the same, but I do not criticize at all your wanting to test it fully.  Let me know if I can help.

cross:
If I could get this in before the next chia-blockchain release, so its in there, that would be ideal.  Is there anything I can do to help testing this?  Just checking since it's been a few weeks here, happy to help if I can.

cross:
Ping.  Sorry to be a noodge, but just wanted to make sure there's a plan to get the required testing done and get this merged.  Again, let me know if I can help.

cross:
@rostislav @hoffmang9 Any status on this PR?

PR #319, opened 2021-12-26
prover_disk: get rid of useless data 

griddis:
while caching C2 to reduce memory usage


dunhil:
Bitblade plotter aligns each plot table by fs block_size
In result cached C2 eats memory quickly.
block_size can be around 1Mb => 260k C2 entries cached, but almost all zeros and usefull

rostislav:
This has been resolved in PR #337.

PR #322, opened 2022-01-11
plotter_disk: Keep streams syncronized and `std::cin` tied

xdustinface:
I think we should drop it since it doesn't bring any value (at least from my understanding) but it messes with the test overview window in CLion. You can run the tests without this PR and as soon as a test created a plot it just doesn't show the result of this and the following tests, instead it shows a endless spinner.

### DIFF
--- a/src/plotter_disk.hpp
+++ b/src/plotter_disk.hpp
@@ -199,9 +199,6 @@ public:
         fs::remove(tmp_2_filename);
         fs::remove(final_filename);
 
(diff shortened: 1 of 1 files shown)


arvidn (approved):

wjblanke:
I don't think we use streams much anymore, but it did improve the performance of the code to drop the sync. Doing this surrounding the plotting code looks fine to me.

PR #323, opened 2022-01-12
src|tests|python-bindings: Make `DiskProver`(de)serializable

xdustinface:
This is a preparation to include `DiskProver` in the `PlotManager` cache in `chia-blockchain` (https://github.com/xdustinface/chia-blockchain/commit/1fdbb4399edd58ccc9c4c5816a364b5a551ea1f4).

### DIFF
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -135,6 +135,17 @@ public:
         deserializer >> C2;
     }
 
+    DiskProver(DiskProver&& other) noexcept
+    {
+        std::lock_guard<std::mutex> lock(other._mtx);
+        filename = std::move(other.filename);
+        memo = std::move(other.memo);
+        id = std::move(other.id);
+        k = other.k;
+        table_begin_pointers = std::move(other.table_begin_pointers);
+        C2 = std::move(other.C2);
+    }
+
     ~DiskProver()
     {
         std::lock_guard<std::mutex> l(_mtx);
@@ -148,7 +159,11 @@ public:
 
     const std::vector<uint8_t>& GetId() { return id; }
 
-    std::string GetFilename() const noexcept { return filename; }
+    const std::vector<uint64_t>& GetTableBeginPointers() const noexcept { return table_begin_pointers; }
+
+    const std::vector<uint64_t>& GetC2() const noexcept { return C2; }
+
+    const std::string& GetFilename() const noexcept { return filename; }
 
     uint8_t GetSize() const noexcept { return k; }
 
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -33,6 +33,7 @@
 #include "calculate_bucket.hpp"
 #include "encoding.hpp"
 #include "entry_sizes.hpp"
+#include "serialize.hpp"
 #include "util.hpp"
 
 struct plot_header {
@@ -48,6 +49,7 @@ struct plot_header {
 // of space, for a given challenge.
 class DiskProver {
 public:
+    static const uint16_t VERSION{1};
     // The constructor opens the file, and reads the contents of the file header. The table pointers
     // will be used to find and seek to all seven tables, at the time of proving.
     explicit DiskProver(const std::string& filename) : id(kIdLen)
@@ -115,6 +117,22 @@ public:
         }
 
         delete[] c2_buf;
+    }
+
+    explicit DiskProver(const std::vector<uint8_t>& vecBytes)
+    {
+        Deserializer deserializer(vecBytes);
+        deserializer >> version;
+        if (version != VERSION) {
+            // TODO: Migrate to new version if we change something related to the data structure
+            throw std::invalid_argument("DiskProver: Invalid version.");
+        }
+        deserializer >> filename;
+        deserializer >> memo;
+        deserializer >> id;
+        deserializer >> k;
+        deserializer >> table_begin_pointers;
+        deserializer >> C2;
     }
 
     ~DiskProver()
@@ -233,7 +251,15 @@ public:
         return full_proof;
     }
 
+    std::vector<uint8_t> ToBytes() const
+    {
+        Serializer serializer;
+        serializer << version << filename << memo << id << k << table_begin_pointers << C2;
+        return serializer.Data();
+    }
+
 private:
+    uint16_t version{VERSION};
     mutable std::mutex _mtx;
     std::string filename;
     std::vector<uint8_t> memo;
--- a/src/serialize.hpp
+++ b/src/serialize.hpp
@@ -1,0 +1,151 @@
+// Copyright 2022 Chia Network Inc
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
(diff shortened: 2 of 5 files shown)


xdustinface:
I had to add d36df94 because the TSAN was red due to an issue in blake3, see the commit message for details.

src/serialize.hpp
```diff
@@ -0,0 +1,151 @@
...
+// limitations under the License.
+
+#ifndef SRC_SERIALIZE_HPP_
+#define SRC_SERIALIZE_HPP_
+
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+template<typename Type> class Serializable{
```
arvidn, line 22: it seems risky to me, to default to making everything serializable by just saving its bytes.
I think this should be constrained to specific types. Never pointers or references for instance.

I think you should add something like this:

```
template<typename Type, std::enable_if_t<std::is_trivial_v<Type>, int> = 0>
class Serializable {
```

src/serialize.hpp
```diff
@@ -0,0 +1,151 @@
...
+        }
+        for (size_t i = 0; i < size; ++i) {
+            *((uint8_t*)&out + i) = in[position + i];
+        }
+        return size;
+    }
+};
+
+template<typename TypeIn, typename TypeOut>
+void Serialize(const TypeIn& in, TypeOut& out)
```
arvidn, line 46: as far as I can tell, `TypeOut` is always `std::vector<uint8_t>&`. Passing anything else here would be a compilation error, wouldn't it?
It seems simpler to just spell out that type in this signature, to get nicer error messages.

src/serialize.hpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    Serializable<TypeIn>::SerializeImpl(in, out);
+}
+
+template<typename TypeIn, typename TypeOut>
+size_t Deserialize(const TypeIn& in, TypeOut& out, const size_t position)
```
arvidn, line 52: similarly, `TypeIn` must be `std::vector<uint8_t> const&` afaict. It should be spelled out here for simplicity

src/serialize.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+class Serializer
+{
+    std::vector<uint8_t> data;
+public:
+    template <typename InputType>
+    friend Serializer& operator <<(Serializer& serializer, const InputType& value) {
+        Serialize(value, serializer.data);
+        return serializer;
+    }
```
arvidn, line 116: I think it would be really nice to constrain this operator as well. If you would add a `static bool specialized = true` to all your `Serializable` classes, you could `std::enable_if<Serializable<InputType>::specialized>` here, to make sure this overload only exists for valid argument types

src/serialize.hpp
```diff
@@ -0,0 +1,151 @@
...
+    for (auto& entry : in) {
+        Serialize(entry, out);
+    }
+}
+
+template<typename TypeIn, typename TypeOut>
+size_t DeserializeContainer(const TypeIn& in, TypeOut& out, const size_t position)
+{
+    size_t elements;
+    size_t size = Deserialize(in, elements, position);
```
arvidn, line 70: I thought I spotted a bug here at first.
`elements` and `size` is a bit confusingly named here, since `size` in most other places refer to the number of elements.
What do you think of renaming `elements` -> `size` and `size` -> `ret` or something like that?

arvidn (commented):
looks good generally. I left a few comments.

There are a few places in the serializer code where I think you should be a bit stricter with which types are accepted. Both constraining with `enable_if` but also just spelling out concrete types.

I would prefer using something like `std::span<uint8_t>` for inputs rather than requiring data to be copied into a `std::vector`, but we don't have a span type in this codebase, and we probably don't want to require that new of a C++ version.

src/prover_disk.hpp
```diff
@@ -117,6 +119,33 @@ class DiskProver {
...
+        }
+        deserializer >> filename;
+        deserializer >> memo;
+        deserializer >> id;
+        deserializer >> k;
+        deserializer >> table_begin_pointers;
+        deserializer >> C2;
+    }
+
+    DiskProver(DiskProver&& other) noexcept
```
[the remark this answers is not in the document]
xdustinface: See cd6806486e7f0aec258520c393ca56acfc0e83f8.

src/prover_disk.hpp
```diff
@@ -117,6 +119,33 @@ class DiskProver {
...
+    {
+        std::lock_guard<std::mutex> lock(other._mtx);
```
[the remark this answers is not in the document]
xdustinface: > It seems unlikely that any other thread could (legitimately) have a reference to this object, since it's an r-value reference.

Good point, see 068feffcc0802c8f563c2f58f9979a7010917501.

src/prover_disk.hpp
```diff
@@ -117,6 +119,33 @@ class DiskProver {
...
+        filename = std::move(other.filename);
+        memo = std::move(other.memo);
+        id = std::move(other.id);
+        k = other.k;
+        table_begin_pointers = std::move(other.table_begin_pointers);
+        C2 = std::move(other.C2);
+    }
```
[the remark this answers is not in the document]
xdustinface: Also moving the `version` now, see 4e88c609175e404baf9425271aa0459a84089fc9. But we can't use the default move constructor here because `std::mutex` isn't movable.

src/serialize.hpp
xdustinface: See 5221580d6de2f5ef31ad00637bc8a845f9ec7c2a.

src/serialize.hpp
xdustinface: Thoughts about 6ac333106f8d9a82d642cb7bba919c0da5b41be8?

src/serialize.hpp
xdustinface: Hm but then we would need to add a specialized version also for all trivial types, no?

src/serialize.hpp
arvidn: Right. The problem with `static_assert` is that you no longer can detect whether `Serializable` is specialized or not.

you could do it like this:
```
template <typename T, typename Dummy = int>
struct Serializable {
    static bool const is_specialized = false;
};

template <typename T>
struct Serializable<T, typename std::enable_if_t<std::is_trivial_v<T>, int>> {
    static bool const is_specialized = true;
    // ...
};

template <>
struct A<std::string, int> {
    static bool const is_specialized = true;
    // ...
};
```

src/serialize.hpp
arvidn: with mt example above, `is_specialized` would be true for all trivial types (as well as `std::vector` and `std::string`).
So you could constrain it like this:

```
    friend Serializer& operator <<(Serializer& serializer, std::enable_if_t<Serializable<InputType>::specialized, InputType> const& value) {
        Serialize(value, serializer.data);
        return serializer;
    }
```

arvidn (approved):

xdustinface:
Rebased after #325 + released the GIL in https://github.com/Chia-Network/chiapos/pull/323/commits/15df7a05e408d87b8783c5eda2b97cc2557cd91e

arvidn (approved):

PR #324, opened 2022-01-13
fix data race in blake3's get_cpu_features()

arvidn:
details posted here: https://github.com/BLAKE3-team/BLAKE3/issues/220


arvidn:
this is a bit more complex than I expected, since both MacOS and ManyLinux lack support for C11 threads functions (specifically, `call_once()`). This could be made to use the posix threading API. It seems MSVC is also lacking support for C11 threads though.

wjblanke (approved):
aok

arvidn:
I'm not sure you should have landed this. CI was still half-red

hoffmang9:
It hasn't been landed yet - just reviewed.

arvidn:
this has been addressed upstream

PR #326, opened 2022-02-09
Fixes incorrect "Approximate N" specification for the C3 table

harold-b:
Fixes incorrect "Approximate N" specification for table the C3 table where it specified 2^k/param_c1 where
it is supposed to be 2^k-2^k/param_c1.


harold-b:
I still have to generate a commit with the HTML version generated, currently the markdown plugin in VSCode is many more changes than needed, even with the specified configuration in the README configured.

rostislav (approved):
Looks good.

PR #327, opened 2022-02-10
Try go bindings

advanderveer:


advanderveer:
Sorry, I was experimenting with the creation of go binding and accidentally pushed this to the wrong remote. I'm sorry if this inadvertently polluted the git history. You may remove all of this if that is still possible.

PR #332, opened 2022-07-02
include picosha2 in verifier.hpp

olivernyc:
I'm running into some build issues locally: `Use of undeclared identifier 'picosha2'`. Could just be an issue with my environment but it seems like `picosha2.hpp` should be included to me.


MumfMeisterT:
What platform are you trying this on?

olivernyc:
This was on macOS using Xcode

PR #337, opened 2022-09-15
Short-circuit C2 table loading at first encounter of lesser C2 entry

harold-b:
Short-circuits C2 table loading at the first encounter of an C2 table entry which is lesser than the previous C2 table entry.

This prevents `DiskProver` from loading dead-space into memory when reading C2.

As there's no specification for the layout of tables in the plot file, but simply a table address in the plot header, plotters may place tables at addresses not immediately following the previous table. 

Bladebit v1 does this since it writes with direct-io enabled and it does not make any effort to ensure the tables appear immediately after another if the previous table size is unaligned. Therefore the may be empty space between tables given the block size of the filesystem of the directory used to export the plot.

See [Bladebit issue #153](https://github.com/Chia-Network/bladebit/issues/153) and [chiapos issue #13084](https://github.com/Chia-Network/chia-blockchain/issues/13084#issuecomment-1248002955)

When said filesystem has a large block size, the empty space between tables will end up being quite large, and `DiskProver` will load the empty space following the C2 entries. This prevents that from happening.

Because C2 entries are sorted in ascending order, we always expect to load a C2 entry of greater value, therefore if a lesser value is encountered it either means this is a corrupted/bad plot file or potentially there might be empty space, in which case it is safe to stop loading the table.

### DIFF
--- a/src/prover_disk.hpp
+++ b/src/prover_disk.hpp
@@ -110,10 +110,21 @@ public:
 
         // The list of C2 entries is small enough to keep in memory. When proving, we can
         // read from disk the C1 and C3 entries.
+        uint64_t prev_c2_f7 = 0;
         auto* c2_buf = new uint8_t[c2_size];
         for (uint32_t i = 0; i < c2_entries - 1; i++) {
             SafeRead(disk_file, c2_buf, c2_size);
-            this->C2.push_back(Bits(c2_buf, c2_size, c2_size * 8).Slice(0, k).GetValue());
+
+            const uint64_t f7 = Bits(c2_buf, c2_size, c2_size * 8).Slice(0, k).GetValue();
+
+            // Short-circuit reading of the C2 table as soon as we encounter an f7 entry whose
+            // value is lesser than the previous f7 read. This ensures that we don't read
+            // empty space from C2 tables that are written with file system block-alignment (like bladebit v1 does).
+            if (f7 < prev_c2_f7)
+                break;
+
+            this->C2.push_back(f7);
+            prev_c2_f7 = f7;
         }
 
         delete[] c2_buf;


xdustinface (dismissed):
Looks good

src/prover_disk.hpp
```diff
@@ -110,10 +110,21 @@ class DiskProver {
...
+        uint64_t prev_c2_f7 = 0;
         auto* c2_buf = new uint8_t[c2_size];
         for (uint32_t i = 0; i < c2_entries - 1; i++) {
             SafeRead(disk_file, c2_buf, c2_size);
-            this->C2.push_back(Bits(c2_buf, c2_size, c2_size * 8).Slice(0, k).GetValue());
+
+            const uint64_t f7 = Bits(c2_buf, c2_size, c2_size * 8).Slice(0, k).GetValue();
+
+            // Short-circuit reading of the C2 table as soon as we encounter an f7 entry whose 
+            // value is lesser than the previous f7 read. This ensures that we don't read
```
rostislav, line 121: Please do not leave trailing whitespace (here and on the previous line).

rostislav (changes requested):
The change itself is okay, but the commit message does not look good. Its first line should fit into 72 characters if possible and be followed by a blank line - this is so that `git log --oneline` output looks pretty. The title of this PR could be used there, with a small change: please use imperative mood ("Short-circuit C2 ..." instead of "Short-circuits C2 ..."). And no extra indentation for the rest of the commit message.

src/prover_disk.hpp
harold-b: Mmm looks like something got butchered when I copied the message from the wrong repo, and pasted it into the correct one. I'll fix it.

xdustinface (approved):

rostislav (approved):
Looks good

PR #341, opened 2022-12-12
Update blake3 to v1.3.3

AmineKhaldi:


cmmarslender:
Blake3 now pulled in with fetch content in cmake, and is at 1.5.0

PR #348, opened 2023-03-06
Remove global variable

nazar-pc:
Generally bad practice to use global variables and easily fixable in this case

### DIFF
--- a/src/phase1.hpp
+++ b/src/phase1.hpp
@@ -47,6 +47,19 @@
 #include "util.hpp"
 #include "progress.hpp"
 
+struct SharedData {
+    uint64_t left_writer_count;
+    uint64_t right_writer_count;
+    uint64_t matches;
+    std::unique_ptr<SortManager> L_sort_manager;
+    std::unique_ptr<SortManager> R_sort_manager;
+    uint64_t left_writer_buf_entries;
+    uint64_t left_writer;
+    uint64_t right_writer;
+    uint64_t stripe_size;
+    uint8_t num_threads;
+};
+
 struct THREADDATA {
     int index;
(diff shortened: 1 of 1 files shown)


arvidn:
I suspect that this will make the code run slower, because of the extra indirections. Is there a stronger argument than "generally bad practice"?

nazar-pc:
I don't think you'll be able to detect any difference in performance here. CPUs are crazy good at these kinds of patterns.

My context is that I was working on turning this CLI into a library and needed to call it potentially many times from the same process, so having global variables was undesirable. There is one more place with static variable in disk reads that would still be a blocker for that though.

I saw the opportunity to improve code a bit and decided to contribute back, but if it is not a welcome change that is fine too.

arvidn:
we have measured performance degradation with smaller changes in the past. If you can demonstrate that there's no slow-down, I think this is a fine change.
I think your use case is a pretty strong argument and might be worth a small slow-down. But I think it would have to be quantified (and not assumed or guessed)

arvidn:
on the other hand, given bladebit, perhaps we don't care so much about a performance degradation in chiapos.

nazar-pc:
I ran `ProofOfSpace` with tmpfs a few times on 13900k and took the best result out of a few runs.

Before:
```
k22 7.786
k25 65.272
```
After:
```
k22 7.752
k25 63.996
```

I'm not sure it became faster, but shouldn't be significantly slower either. My test was not super scientific, I have a bunch of things running in background on my machine.

MumfMeisterT:
We're short on time to review this, but if you fork chiapos and it works well for you, we'll consider up-streaming it.

PR #358, opened 2023-05-05
Cleanup work from fc.proving_cuda

harold-b:


harold-b:
This needs to be merged via "Squash and merge" to `main-bb-prep`

PR #383, opened 2023-08-16
build wheels for macOS 10.14 for now

altendky:
https://github.com/Chia-Network/chia-blockchain/pull/16047#pullrequestreview-1579705035


emlowe:
We seem to need another chiapos release before 2.0.0 so this is important to fix up

altendky:
As far as we can tell so far, TSAN failures are separate and are being explored.

emlowe:
Close and re-open for new CI

PR #386, opened 2023-08-16
Use ubuntu 22 runners

emlowe:


arvidn (approved):

altendky (dismissed):
isn't this a debug?

emlowe:
I suggest we just switch right now, since I'm convinced the problem is a runner issue, and whether we run this on 20 or 22 doesn't matter

altendky:
Alrighty, I think I'm convinced that while this is a questionable test of our artifacts (since it doesn't test them nor make any attempt to match their build environment) that changing from 20.04 to 22.04 isn't particularly worse.

PR #394, opened 2023-09-01
Add riscv64 info and simplify macos

emlowe:
Simplify macos checks by just using CMAKE_SYSTME_PROCESSOR and add in riscv64

### DIFF
--- a/.github/workflows/build-test-riscv64.yml
+++ b/.github/workflows/build-test-riscv64.yml
@@ -47,7 +47,5 @@ jobs:
             mkdir -p build && \
             cd build && \
             cmake ../ && \
-            cmake --build . -- -j 2 && \
-            echo "Running RunTests" && \
-            ./RunTests
+            cmake --build . -- -j 2
           '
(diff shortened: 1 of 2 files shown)


hoffmang9 (dismissed):
lgtm

.github/workflows/build-test-riscv64.yml
```diff
@@ -0,0 +1,59 @@
...
+      if: startsWith(matrix.os, 'ubuntu-latest')
+      id: qemu
+      uses: docker/setup-qemu-action@v2
+      with:
+        platforms: riscv64
+#        image: tonistiigi/binfmt:latest
+
+    - name: Build and Test
+      run: |
+        docker run --rm \
```
cmmarslender, line 39:
```suggestion
        docker run --rm --platform linux/riscv64 \
```

I think you get it implicitly because of the image, but usually prefer to specify `--platform linux/riscv64` as well

cmmarslender (approved):

hoffmang9:
BTW: As a person not as conversant in docker as I'd like to be - there may be a better riscv64/debian image and/or it might make more sense for us to cache on that has more python and build tools - but also don't want to make any more work than this fun side project.

cmmarslender (approved):

PR #396, opened 2023-09-05
Actually run the tests on Windows using `RunTests`

emlowe:
Windows was running the tests with `ctest` - this actually wasn't running any tests:

```
Test project D:/a/chiapos/chiapos/build-win
No tests were found!!!
```

### DIFF
--- a/.github/workflows/build-test-cplusplus.yml
+++ b/.github/workflows/build-test-cplusplus.yml
@@ -82,4 +82,4 @@ jobs:
         cd build-win
         cmake ..
         cmake --build . --config Release -j 6
-        ./RunTests
+        ./RunTests.exe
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-cplusplus.yml
```diff
@@ -82,4 +82,4 @@ jobs:
         cd build-win
         cmake ..
         cmake --build . --config Release -j 6
-        ctest -C Release -j 6
+        ./RunTests
```
hoffmang9, line 85: RunTests.exe

hoffmang9 (approved):
Those darn slashes

emlowe:
Needed the correct location too - was in the Release directory and not the current one. I got it on the 3rd try though  :-)

cmmarslender:
The original command (`ctest -C Release -j 6`) works again now after https://github.com/Chia-Network/chiapos/pull/402 was merged. It was broken when the compressed harvesting PR merged and had commented out the last couple of lines in the cmake file that disabled adding the tests.

Is there a benefit to running the tests this way vs using `ctest` now that `ctest` works again?

PR #402, opened 2023-10-13
Swap blake3 to FetchContent + general cleanup

cmmarslender:
- Gets blake3 via fetchcontent in cmake
- Builds windows wheels with cmake as well, vs the old split method used in setup.py
- re-enables the tests on windows which were disabled when bladebit compress harvesting was merged
- Fix edge case in C++ tests on mac and add mac C++ tests to CI


hoffmang9 (commented):
My only question is - now that we're not specifying the assembly versions, will Windows still get the CPU optimized code?

cmmarslender:
> My only question is - now that we're not specifying the assembly versions, will Windows still get the CPU optimized code?

it appears they handle this all on their end now (they added cmake config since we initially pulled in the files).
For instance: https://github.com/BLAKE3-team/BLAKE3/blob/master/c/CMakeLists.txt#L74:L81

Its easier to see the output from their cmake file in the c++ build/test workflows than the wheels, but as another example, they have this output that shows up when assembly is enabled: https://github.com/BLAKE3-team/BLAKE3/blob/master/c/CMakeLists.txt#L180

Which we can see in our windows build here: https://github.com/Chia-Network/chiapos/actions/runs/6594714279/job/17918903037?pr=402#step:3:40

hoffmang9 (approved):
I now remember Amine making those changes.

PR #407, opened 2024-01-23
Add python versions and upload to chia pypi for riscv workflow

emlowe:
Matches the chiavdf workflow and adds in the python versions and an upload step


hoffmang9 (approved):
lgtm

PR #408, opened 2024-01-23
Remove cmake version check from setup.py

emlowe:
There is a cmake version is already specified in `CMakeLists.txt` and this Windows-specific check in setup.py isn't required (and wasn't accurate anyway)


hoffmang9 (approved):
lgtm

PR #411, opened 2024-01-23
Remove small runner workarounds

emlowe:
See if the larger default runners allow us to remove some of these small runner hacks


hoffmang9 (approved):
Nice catch

PR #439, opened 2024-08-12
CHIA-193: Rust bindings to chiapos

Rigidity:
Adds a Rust crate with a memory safe (at least, according to fuzz tests and unit tests I've written) binding to `validate_proof`. This is part of an effort to port block prevalidation to Rust.


CMakeLists.txt
```diff
@@ -156,7 +156,15 @@ include_directories(
...
 IF (BUILD_PROOF_OF_SPACE_STATICALLY)
   message("Statically build ProofOfSpace")
-  target_link_libraries(ProofOfSpace -static -Wl,--whole-archive -lrt -lpthread -Wl,--no-whole-archive)
+  target_link_libraries(ProofOfSpace PUBLIC -static -Wl,--whole-archive -lrt -lpthread -Wl,--no-whole-archive)
+ENDIF()
+
+option(BUILD_STATIC_CHIAPOS_LIBRARY "Build chiapos library statically" OFF)
+IF (BUILD_STATIC_CHIAPOS_LIBRARY)
+  message("Statically build chiapos library")
+  add_library(chiapos_static STATIC src/chacha8.c src/verifier.hpp c-bindings/wrapper.cpp)
```
arvidn, line 165: you're adding a `.hpp` file as a source file here. That looks a bit suspicious. If it's correct, I think it warrants a comment. Does `wrapper.cpp` not include this header file?

CMakeLists.txt
```diff
@@ -156,7 +156,15 @@ include_directories(
 option(BUILD_PROOF_OF_SPACE_STATICALLY "Build ProofOfSpace target statically" OFF)
 IF (BUILD_PROOF_OF_SPACE_STATICALLY)
   message("Statically build ProofOfSpace")
-  target_link_libraries(ProofOfSpace -static -Wl,--whole-archive -lrt -lpthread -Wl,--no-whole-archive)
+  target_link_libraries(ProofOfSpace PUBLIC -static -Wl,--whole-archive -lrt -lpthread -Wl,--no-whole-archive)
+ENDIF()
+
+option(BUILD_STATIC_CHIAPOS_LIBRARY "Build chiapos library statically" OFF)
```
arvidn, line 162:
```suggestion
option(BUILD_STATIC_CHIAPOS_LIBRARY "Build chiapos static library (verify-only)" OFF)
```

Nit pick, a library is "static" and you can link against it "statically".

c-bindings/wrapper.cpp
```diff
@@ -0,0 +1,19 @@
...
+    ByteArray validate_proof(const uint8_t* seed, uint8_t k, const uint8_t* challenge, const uint8_t* proof, uint16_t proof_len) {
+        Verifier v;
+        auto quality = v.ValidateProof(seed, k, challenge, proof, proof_len);
+        if (quality.GetSize() == 0) {
+            return ByteArray { nullptr, 0 };
+        }
+        uint8_t *quality_buf = new uint8_t[32];
+        quality.ToBytes(quality_buf);
+        delete[] quality_buf;
+        return ByteArray { quality_buf, 32 };
```
arvidn, line 13: I think it would be a lot less error prone to take a pointer to a 32 byte buffer where you copy the quality, rather than heap allocating a buffer that must be freed later. You could return a `bool` indicating success.

c-bindings/wrapper.h
```diff
@@ -0,0 +1,13 @@
+#include "picosha2.hpp"
+#include "../src/verifier.hpp"
+
+extern "C" {
+    typedef struct {
+        uint8_t* data;
+        size_t length;
+    } ByteArray;
```
arvidn, line 8: with my suggestion you wouldn't need this type. I think it's fine to have a C interface where the caller must know about the (fixed) size of buffers. You immediately wrap it in a rust binding where this information is encoded in the type anyway.

c-bindings/wrapper.h
```diff
@@ -0,0 +1,13 @@
+#include "picosha2.hpp"
+#include "../src/verifier.hpp"
```
arvidn, line 2: I would remove this include here and move it into `wrapper.cpp`. This interface doesn't depend on the verifier class itself.
Also, it would be best to not have a relative path here, it's something compilers don't always like (being able to "escape" include directories)

arvidn (commented):
looks good, just a few comments

c-bindings/wrapper.cpp
```diff
@@ -1,19 +1,14 @@
 #include "wrapper.h"
+#include "verifier.hpp"
 
 extern "C" {
-    ByteArray validate_proof(const uint8_t* seed, uint8_t k, const uint8_t* challenge, const uint8_t* proof, uint16_t proof_len) {
+    bool validate_proof(uint8_t* seed, uint8_t k, uint8_t* challenge, uint8_t* proof, uint16_t proof_len, uint8_t* quality_buf) {
```
arvidn, line 5: I see you removed the `const`s here. You aren't mutating those arrays, are you?

c-bindings/wrapper.h
```diff
@@ -1,13 +1,5 @@
...
 extern "C" {
-    typedef struct {
-        uint8_t* data;
-        size_t length;
-    } ByteArray;
-
-    ByteArray validate_proof(const uint8_t* seed, uint8_t k, const uint8_t* challenge, const uint8_t* proof, uint16_t proof_len);
-
-    void delete_byte_array(ByteArray array);
+    bool validate_proof(uint8_t* seed, uint8_t k, uint8_t* challenge, uint8_t* proof, uint16_t proof_len, uint8_t* quality_buf);
```
arvidn, line 4: I think removing the `const`s was a mistake here.

rust-bindings/src/lib.rs
```diff
@@ -21,22 +21,14 @@ pub fn validate_proof(
...
     unsafe {
-        let array = bindings::validate_proof(
-            seed.as_ptr(),
+        bindings::validate_proof(
+            seed.as_ptr() as *mut u8,
             k,
-            challenge.as_ptr(),
-            proof.as_ptr(),
+            challenge.as_ptr() as *mut u8,
+            proof.as_ptr() as *mut u8,
```
arvidn, line 28: I think making these mutable was a mistake

arvidn (approved):

rust-bindings/src/lib.rs
```diff
@@ -36,6 +36,34 @@ pub fn validate_proof(
 mod tests {
     use super::*;
 
+    #[test]
+    fn test_proofs() {
+        /*
+         * Generated by printing the inputs to validate_proof in tests/test_python_bindings.py
+         * Specifically, the test_k_21 function was adapted to append 100 lines to the file.
+         */
+        let proofs = include_str!("../test_proofs.txt");
```
arvidn, line 45: not important; but I would think it would be faster to compile to read this file at run-time rather than compile time

rust-bindings/src/lib.rs
```diff
@@ -36,6 +36,34 @@ pub fn validate_proof(
...
+
+            let mut actual_quality = [0; 32];
+            assert!(validate_proof(
+                &seed.try_into().unwrap(),
+                k,
+                &challenge.try_into().unwrap(),
+                &proof,
+                &mut actual_quality
+            ));
+            assert_eq!(actual_quality.as_slice(), quality);
```
arvidn, line 63: it might make sense to also ensure that `validate_proof()` *fails* if you corrupt any of the inputs.

rust-bindings/test_proofs.txt
```diff
@@ -0,0 +1,97 @@
+05683404333717545b0a6f0c0dde9710e4d3fe2d5cc6cc0a090a0b818bab0f17, 21, df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119, cc128d31890be2ef01ed47503b8559caa2014b9262651db7c43ff1544acb0de1f354e87d88b0b1c87383394d52b6fc393fbacb0224665488cd31d7c6f52d4113a7d3e3c5cc5366668dd446586bc9c66fb607ea55596ce2d14db5ff964c2f7c7ffc4f5590b1313c6ce29a2b0ef6773256d3019f0a78ace7aef2efd8c2decefc317a98bd61722839c1b68d28eb38972b4b12b5912d9024e9b8d5a4cd878bc12192b1390a7d9dd6b82a, 753087568382550fdefa85d7907efd91d7ad10a6b2e4787cf97f848f4753302e
```
arvidn, line 1: do you think it makes sense to also include the script you used to generate this file? just in case we want to extend it or modify it in the future.

tests/test.cpp
```diff
@@ -598,16 +598,18 @@ TEST_CASE("K Sizes")
 
     SECTION("Minimum K Size")
     {
-        uint8_t proof_data[0];
+        uint8_t *proof_data = new uint8_t[0];
```
arvidn, line 601: you're not allowed to allocate 0-length arrays. Both this and the previous version are invalid.
Don't you know the size of this buffer?
You could probably just do something like:
```
uint8_t proof_data[32];
```
and pass a pointer to that.

arvidn (commented):
looks good to me. do you want Harold to have time to review it as well?

Rigidity:
> looks good to me. do you want Harold to have time to review it as well?

Sounds like a good idea if he's able to, I'm not in a rush to release this but this week would be nice

tools/generate_rust_proof_test.py
```diff
@@ -0,0 +1,44 @@
...
+os.remove(path)
+
+v = Verifier()
+for i in range(100):
+    if i % 100 == 0:
+        print(i)
+    challenge = sha256(i.to_bytes(4, "big")).digest()
+    for index, quality in enumerate(pr.get_qualities_for_challenge(challenge)):
+        proof = pr.get_full_proof(challenge, index)
+        assert len(proof) == 8 * pr.get_size()
```
haorldbchi, line 38: Note that this *may* fail. Although most proofs will be present more often than not, it can still fail. It should be good enough to try again until a proof comes back.

rust-bindings/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+#![allow(non_upper_case_globals)]
+
+extern crate link_cplusplus;
+
+mod bindings {
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+}
+
+pub fn validate_proof(
+    seed: &[u8; 32],
```
haorldbchi, line 13: In all the code and literature this is referred to as "plot_id" or "id", it think the same nomenclature should be retained here.

c-bindings/wrapper.cpp
```diff
@@ -0,0 +1,14 @@
+#include "wrapper.h"
+#include "verifier.hpp"
+
+extern "C" {
+    bool validate_proof(const uint8_t* seed, uint8_t k, const uint8_t* challenge, const uint8_t* proof, uint16_t proof_len, uint8_t* quality_buf) {
+        Verifier v;
+        auto quality = v.ValidateProof(seed, k, challenge, proof, proof_len);
+        if (quality.GetSize() == 0) {
```
haorldbchi, line 8: Since we're already checking for size here, perhaps we can be paranoid and instead return if  `quality.GetSize() != 256`.

c-bindings/wrapper.h
```diff
@@ -0,0 +1,5 @@
+#include "picosha2.hpp"
+
+extern "C" {
+    bool validate_proof(const uint8_t* seed, uint8_t k, const uint8_t* challenge, const uint8_t* proof, uint16_t proof_len, uint8_t* quality_buf);
```
haorldbchi, line 4: Same as in `lib.rs` (`seed` vs `id` nomenclature).

haorldbchi (changes requested):

tools/generate_rust_proof_test.py
haorldbchi: Good catch, my mistake.

c-bindings/wrapper.cpp
```diff
@@ -2,10 +2,10 @@
 #include "verifier.hpp"
 
 extern "C" {
-    bool validate_proof(const uint8_t* seed, uint8_t k, const uint8_t* challenge, const uint8_t* proof, uint16_t proof_len, uint8_t* quality_buf) {
+    bool validate_proof(const uint8_t* plot_id, uint8_t k, const uint8_t* challenge, const uint8_t* proof, uint16_t proof_len, uint8_t* quality_buf) {
         Verifier v;
-        auto quality = v.ValidateProof(seed, k, challenge, proof, proof_len);
-        if (quality.GetSize() == 0) {
+        auto quality = v.ValidateProof(plot_id, k, challenge, proof, proof_len);
+        if (quality.GetSize() != 256) {
```
arvidn, line 8: is it possible to have *fewer* than 256 bits (32 bytes) of quality? If so, you may need to print leading zeros into `quality_buf`

arvidn (commented):
good call on the `GetSize()` check

arvidn (approved):

c-bindings/wrapper.cpp
harold-b: According to https://github.com/Chia-Network/chiapos/blob/e5eeb6ed591cf22f67ad528d04ace2d8e635ee5f/src/verifier.hpp#L55 it should always return either exactly 32 bytes/256 bits or 0

haorldbchi (approved):

PR #442, opened 2024-08-21
Allow offline builds of chiapos with local repos

Rigidity:


arvidn (commented):
two aspects make me a bit uneasy about this:
* whether you enable `BUILD_OFFLINE` or not depends on whether the `thirdparty` directory exists or not. That's an unexpected (and unintuitive, I would say) mechanism to communicate with the build script.
* the commit hashes of the dependencies are repeated in two places. I can see a future PR where we bump the versions, but forget one of those places.

I'm not sure what the best approach is, in a way I think submodules would be a pretty good alternative to cmake fetchcontent, but I'm not familiar enough with cmake to really recommend that. It might cause a bunch of trouble. But I think it's fairly standard practice in the C++ world to pull in open-source dependencies that way.

Perhaps the simplest approach would be to, essentially, replace the `install-dependencies.sh` with a call to `cmake` but to just download dependencies.

hoffmang9:
Agreeing with Arvid that the build process should work the same whether building locally or on CI. I think we're both missing something about the motivation here?

arvidn:
The discussion happened on keybase yesterday. The summary is:

* rust creates don't allow the build step to pull things from the internet. I suspect the main reason is to ensure builds are deterministic.
* our cmake file uses FetchContent to download dependencies, which means we building the rust bindings fail
* We want a way to download and bundle the dependencies in the crate separate from the building step

PR #450, opened 2024-09-11
Just use `std::filesystem`

emlowe:
We require C++17 where `std::filesystem` is standard and we don't support some of the older systems (like Mojave) anymore where we need to use some other things like `gulrak` or `experimental`


wjblanke (approved):
aok

arvidn (approved):

PR #456, opened 2024-10-30
Try setting specific output directory for static libs

emlowe:


emlowe:
@AmineKhaldi Do you see a downside to this - it makes building the rust bindings simpler since all the libs will be in a single directory on all the platforms (I think anyway)

AmineKhaldi:
> @AmineKhaldi Do you see a downside to this - it makes building the rust bindings simpler since all the libs will be in a single directory on all the platforms (I think anyway)

Hmm this doesn't make sense to me, AFAIK genex do not expand at that time, and they're mainly meant to be used with targets, so if this is working like this, it seems unintended.

Another point to consider is that by setting this here, it will apply to everything below it. I've sent an experiment that targets specifically what we need, please let me know if it works: https://github.com/Chia-Network/chiapos/pull/458

PR #470, opened 2025-01-31
don't access proof after dropping gil

emlowe:
In `validate_proof` the `proof` variable is accessed after dropping the GIL. This can result in crashes as python may have released that memory at any time after the GIL is dropped.

Make sure to use a temp var to hold the proof size so we can avoid referencing this var


hoffmang9 (approved):
lgtm

hoffmang9 (approved):
lgtm again

arvidn (approved):
just a nit: the `proof_size` variable is unnecessary

emlowe:
close and reopen to pick up needed workflow changes

hoffmang9 (approved):
lgtm unless you want Arvid's suggestion

PR #475, opened 2025-02-06
Update bladebit harvester artifact version for linux x86-64

haorldbchi:


emlowe:
The new bladebit libs used here (the linux v3.1.1 versions) were built with a newer glibc then we currently use to build chiapos - so they won't work. Moving this PR to draft for now.

PR #476, opened 2025-02-06
Manually clear GNU_STACK executable bit on bladebit libraries

emlowe:
Manually clear using `execstack -c` the GNU_STACK executable bit on the bladebit dynamic libraries after fetching the artifacts.

(also added `--ssl-revoke-best-effort` for windows to avoid bizarre CRL issues in CI. Probably some GH issue really)

### DIFF
--- a/.github/actions/fetch_bladebit_harvester.sh
+++ b/.github/actions/fetch_bladebit_harvester.sh
@@ -89,7 +89,7 @@ if [[ "${artifact_ext}" == "zip" ]]; then
 else
   pushd "${dst_dir}"
   tar -xzvf "../../${artifact_name}"
-  if [[ "${host_os}" == "linux" ]] && [[${host_arch}" == "x86-64]]; then
+  if [[ "${host_os}" == "linux" ]] && [[ "${host_arch}" == "x86-64" ]]; then
     execstack -c lib/libbladebit_harvester.so
   fi
   popd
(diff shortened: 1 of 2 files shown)


haorldbchi:
Testing decompression via `ProofOfSpace check` on a compressed plot with libbladebit_harvestor.so having its `GNU_STACK` executable bit ripped worked fine in a C1 plot:

```
Proof verification succeeded. k = 32
Total success: 1007/1000, 100.7%.
Total failures: 0
Exceptions: 0

readelf -e libbladebit_harvester.so | grep -A 1 GNU_STACK

  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x10
````

Both CPU and GPU paths were tested yielding the same results.

**Review by arvidn — DISMISSED** [review]

File: .github/actions/fetch_bladebit_harvester.sh [hunk 1]
```diff
@@ -89,5 +91,9 @@ if [[ "${artifact_ext}" == "zip" ]]; then
 else
   pushd "${dst_dir}"
   tar -xzvf "../../${artifact_name}"
+  if [[ "${host_os}" == "linux" ]] && [[ "${host_arch}" == "x86-64" ]]; then
+    # On linux clear the GNU_STACK executable bit for glibc 2.41 compat
+    execstack -c lib/libbladebit_harvester.so
+  fi
```
altendky, line 97: i think there ought to be a note explaining why this is needed and at what point it should be removed.  should have a `# TODO:`.

altendky (dismissed):

PR #477, opened 2025-02-09
update pybind and fix race for bucket init. also stress test

wjblanke:
DO NOT MERGE... yet

### DIFF
--- a/tests/POSStressTest.cpp
+++ b/tests/POSStressTest.cpp
@@ -1,0 +1,88 @@
+#include <stdio.h>
+#include "verifier.hpp"
+
+#include <sstream>
+#include <string>
+#include <fstream>
+#include <thread>
+
+std::vector<uint8_t> HexToBytes(const char *hex_proof) {
+    int len = strlen(hex_proof);
+    assert(len % 2 == 0);
+    std::vector<uint8_t> result;
+    for (int i = 0; i < len; i += 2)
+    {
+        int hex1 = hex_proof[i] >= 'a' ? (hex_proof[i] - 'a' + 10) : (hex_proof[i] - '0');
+        int hex2 = hex_proof[i + 1] >= 'a' ? (hex_proof[i + 1] - 'a' + 10) : (hex_proof[i + 1] - '0');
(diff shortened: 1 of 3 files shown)


tests/POSStressTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+        int hex1 = hex_proof[i] >= 'a' ? (hex_proof[i] - 'a' + 10) : (hex_proof[i] - '0');
+        int hex2 = hex_proof[i + 1] >= 'a' ? (hex_proof[i + 1] - 'a' + 10) : (hex_proof[i + 1] - '0');
+        result.push_back(hex1 * 16 + hex2);
+    }
+    return result;
+}
+
+void doit(int thread)
+{
+    std::ifstream infile("pos.txt");
```
arvidn, line 24: I think there's a risk this won't be as "stressing" is it looks, since every thread will sleep in disk I/O in every iteration of the loop. I would think it would put a lot more load (and stress) by loading and parsing the file once in the main thread, and have the worker threads just read from there

tests/POSStressTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+    std::thread bunch0(doit,0);
+    std::thread bunch1(doit,1);
+    std::thread bunch2(doit,2);
+    std::thread bunch3(doit,3);
+    std::thread bunch4(doit,4);
+    std::thread bunch5(doit,5);
+    std::thread bunch6(doit,6);
+    std::thread bunch7(doit,7);
+    std::thread bunch8(doit,8);
```
arvidn, line 74: a bit of a nit, but I think this should be a loop, and we might want more than 10 threads

arvidn (approved):
This is definitely a data race, so a good fix.
More generally, keeping global variables in a shared library is really fishy. including the mutex itself. Sometimes unexpected things happen if libraries are unloaded and reloaded but the data section being preserved. I don't think we would get in trouble for this, it's just smelly.
It would probably be better to compute this table at compile time (or as a separate pre-processing step perhaps).

